PR #1, opened 2020-10-27
Review

notgne2:


src/main.rs
```diff
@@ -0,0 +1,439 @@
...
+
+    if !check_status.success() {
+        good_panic!("Checks failed for the given flake repo");
+    }
+
+    ()
+}
+
+/// Evaluates the Nix in the given `repo` and return the processed Data from it
+async fn get_deployment_data(
```
balsoft, line 243: Maybe consider only partially evaluating the `deploy`, so that we don't have to wait any longer than we have to

src/utils/deploy.rs
```diff
@@ -0,0 +1,169 @@
...
+        "Activating profile `{}` for node `{}`",
+        deploy_data.profile_name, deploy_data.node_name
+    );
+
+    let activate_path_str = super::deploy_path_to_activate_path_str(&deploy_defs.current_exe)?;
+
+    let temp_path: Cow<str> = match &deploy_data.merged_settings.temp_path {
+        Some(x) => x.into(),
+        None => "/tmp/deploy-rs".into(),
+    };
```
balsoft, line 89:
```suggestion
    let temp_path: Cow<str> = deploy_data.merged_settings.temp_path.as_ref().unwrap_or("/tmp/deploy-rs").into();
```

(Untested, don't just commit it in)

src/utils/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+}
+
+#[derive(PartialEq, Debug)]
+pub struct DeployFlake<'a> {
+    pub repo: &'a str,
+    pub node: Option<&'a str>,
+    pub profile: Option<&'a str>,
+}
+
+pub fn parse_flake(flake: &str) -> DeployFlake {
```
balsoft, line 41: Does Nix parse flakes in the same way? If it does, how does it handle paths with `#` ?

src/utils/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+    pub ssh_user: Cow<'a, str>,
+    pub profile_user: Cow<'a, str>,
+    pub profile_path: Cow<'a, str>,
+    pub current_exe: PathBuf,
+    pub sudo: Option<String>,
+}
+
+impl<'a> DeployData<'a> {
+    pub fn defs(&'a self) -> DeployDefs<'a> {
+        let ssh_user: Cow<str> = match self.merged_settings.ssh_user {
```
balsoft, line 117: `unwrap_or`

src/utils/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+                _ => format!(
+                    "/nix/var/nix/profiles/per-user/{}/{}",
+                    profile_user, self.profile_name
+                )
+                .into(),
+            },
+            Some(ref x) => x.into(),
+        };
+
+        let sudo: Option<String> = match self.merged_settings.user {
```
balsoft, line 146: `map`

src/utils/mod.rs
```diff
@@ -0,0 +1,233 @@
...
+                Some(ref x) => x.into(),
+                None => good_panic!(
+                    "Neither user nor sshUser set for profile `{}` of node `{}`",
+                    self.profile_name,
+                    self.node_name
+                ),
+            },
+        };
+
+        let profile_path: Cow<str> = match self.profile.profile_settings.profile_path {
```
balsoft, line 134: Chained `or_else` and `unwrap_or`

src/utils/push.rs
```diff
@@ -0,0 +1,137 @@
...
+    } else {
+        build_c
+            .arg(&repo)
+            .arg("--no-out-link")
+            .arg("-A")
+            .arg(format!(
+                "deploy.nodes.{}.profiles.{}.path",
+                deploy_data.node_name, deploy_data.profile_name
+            ))
+    };
```
balsoft, line 42: Is there a reason for two separate if's? Seems less readable to me.

src/utils/push.rs
```diff
@@ -0,0 +1,137 @@
...
+
+    build_command = match (keep_result, supports_flakes) {
+        (true, _) => {
+            let result_path = match result_path {
```
balsoft, line 46: `unwrap_or`

balsoft (commented):
Whew.

Two important comments.

1. Please make all commands called by deploy-rs and activate pass their output through to the user!

2. See the comment about bootstrap paths, the logic appears wrong.

The rest is just my smol functional brain disliking imperative programming.

src/activate.rs
```diff
@@ -0,0 +1,303 @@
...
+        fork::Fork::Child => {
+            std::thread::spawn(move || {
+                let mut rt = tokio::runtime::Runtime::new().unwrap();
+
+                rt.block_on(async move {
+                    info!("Waiting for confirmation event...");
+
+                    let mut buffer = [0; 32];
+                    let mut stream =
+                        deactivate_on_err(&profile_path, inotify.event_stream(&mut buffer)).await;
```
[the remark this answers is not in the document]
balsoft: Oh, hm, I guess it fails before even forking off. Might be worth rolling back if that happens.

notgne2:
@balsoft I haven't tested bootstrap at all and I'm not really sure about the purpose so it makes sense that it might not all be logically sound. What exactly is the use case?

src/activate.rs
```diff
@@ -0,0 +1,303 @@
...
+
+async fn deactivate_on_err<A, B: core::fmt::Debug>(profile_path: &str, r: Result<A, B>) -> A {
+    match r {
+        Ok(x) => x,
+        Err(err) => {
+            error!("Deactivating due to error: {:?}", err);
+            match deactivate(profile_path).await {
+                Ok(_) => (),
+                Err(err) => {
+                    error!("Error de-activating, uh-oh: {:?}", err);
```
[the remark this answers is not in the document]
notgne2: I'm not sure what you mean, `error!()` is a logging macro just like `info!()`, and this will print a debug-formatted copy of the error too

src/activate.rs
```diff
@@ -0,0 +1,303 @@
...
+        match bootstrap_status {
+            Ok(s) if s.success() => (),
+            _ => {
+                tokio::fs::remove_file(&profile_path).await?;
+                good_panic!("Failed to execute bootstrap command");
+            }
+        }
+    }
+
+    let activate_status = Command::new(format!("{}/deploy-rs-activate", profile_path))
```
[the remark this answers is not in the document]
notgne2: They are forwarded by default, you have to manually mute the outputs

src/main.rs
```diff
@@ -0,0 +1,439 @@
...
+
+    // Add any profiles which weren't in the provided order list
+    for profile_name in node.node_settings.profiles.keys() {
+        if !profiles_list.contains(&profile_name.as_str()) {
+            profiles_list.push(&profile_name);
+        }
+    }
+
+    for profile_name in profiles_list {
+        let profile = match node.node_settings.profiles.get(profile_name) {
```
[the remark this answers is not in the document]
notgne2: the latter is definitely the most preferable to me, there's just a little more engineering involved so I hadn't gotten around to it yet (I generally see using `Box<dyn std::error::Error>` without a good reason as a hack to make programs compile for prototyping sake)

src/activate.rs
```diff
@@ -0,0 +1,303 @@
...
+                                &profile_path,
+                                timeout(Duration::from_secs(confirm_timeout as u64), stream.next())
+                                    .await,
+                            )
+                            .await
+                            .ok_or("Watcher ended prematurely"),
+                        )
+                        .await,
+                    )
+                    .await;
```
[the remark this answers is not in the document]
notgne2: When I get the error types sorted out this should stop being so ugly (as long as I don't get caught up in move semantics moving the logic to a "danger zone")

src/utils/push.rs
notgne2: Kindof? I can't let the Command itself go out of scope and this solved it

notgne2:
13 panics left to go, and almost every error is a proper one

notgne2:
2 panics left now, dealing with the awful `deactivate_on_err`-inside-fork spaghetti

src/activate.rs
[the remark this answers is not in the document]
notgne2: I think it should have auto rolled back at that point anyway (it had a deactivate on failure wrapper higher up, the only reason it is special lower down is due to the forking, threading, and tokio runtime), but I've rewritten it now and it still should

src/activate.rs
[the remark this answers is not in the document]
notgne2: fixed lol

src/main.rs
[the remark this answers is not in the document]
notgne2: well, this awfully tiring task has been done now!

examples/system/README.md
```diff
@@ -0,0 +1,16 @@
...
+-->
+
+# Example nixos system deployment
+
+This is an example of how to deploy a full nixos system with a separate user unit to a bare machine.
+
+1. Run bare system from `.#nixosConfigurations.bare`
+  - `nix build .#nixosConfigurations.bare.config.system.build.vm`
+  - `QEMU_NET_OPTS=hostfwd=tcp::2221-:22 ./result/bin/run-bare-system-vm`
+2. `nix run github:serokell/deploy --prime`
```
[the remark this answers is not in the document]
notgne2: I've removed `--prime` and I think everything else works

flake.nix
```diff
@@ -0,0 +1,76 @@
...
+            schema = deploy: pkgs.runCommandNoCC "jsonschema-deploy-system" { } ''
+              ${pkgs.python3.pkgs.jsonschema}/bin/jsonschema -i ${pkgs.writeText "deploy.json" (builtins.toJSON deploy)} ${./interface/deploy.json} && touch $out
+            '';
+
+            activate = deploy:
+              let
+                allPaths = pkgs.lib.flatten (pkgs.lib.mapAttrsToList (nodeName: node: pkgs.lib.mapAttrsToList (profileName: profile: profile.path) node.profiles) deploy.nodes);
+              in
+              pkgs.runCommandNoCC "deploy-rs-check-activate" { } ''
+                for i in ${builtins.concatStringsSep " " allPaths}; do test -f "$i/deploy-rs-activate" || (echo "A profile path is missing an activation script" && exit 1); done
```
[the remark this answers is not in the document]
notgne2: Done!

src/activate.rs
```diff
@@ -0,0 +1,303 @@
...
+        .arg(&closure)
+        .stdout(Stdio::null())
+        .status()
+        .await?;
+
+    if !nix_env_set_exit_status.success() {
+        good_panic!("Failed to update nix-env generation");
+    }
+
+    if let (Some(bootstrap_cmd), false) = (bootstrap_cmd, !Path::new(&profile_path).exists()) {
```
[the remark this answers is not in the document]
notgne2: See my comment on bootstrap

src/utils/mod.rs
notgne2: I'm not sure what you mean, the flake URL given to `nix build` and the one given to `deploy` will already be very different as stuff like `.deploy` `.nodes`, .`profiles`, and `.path` are left out

src/main.rs
```diff
@@ -0,0 +1,439 @@
...
+    info!("Pushing all profiles for `{}`", node_name);
+
+    let mut profiles_list: Vec<&str> = node
+        .node_settings
+        .profiles_order
+        .iter()
+        .map(|x| x.as_ref())
+        .collect();
+
+    // Add any profiles which weren't in the provided order list
```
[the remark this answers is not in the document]
notgne2: This is specifically inside `push_all_profiles` which is only called if you run `deploy` or `deploy .#node`

src/main.rs
notgne2: The evaluation time is certainly annoying but I think it's time I take a break from making so many major changes, if somebody else could do it that would be ideal

src/utils/mod.rs
notgne2: More like
```rust
let ssh_user: Cow<str> = self
    .merged_settings
    .ssh_user
    .as_deref()
    .map(|x| x.into())
    .unwrap_or_else(|| whoami::username().into());
```
which I think is a little uglier, and longer

src/utils/mod.rs
notgne2: There's a guard pattern in use too, so the functional replacement would be `and_then` not `map` (as the guard would have to be replaced with a ternary/if expression so it can be `None` if the condition is not met).

In the end it would look like this which I don't think is much better
```rust
let sudo: Option<String> = self.merged_settings.user.as_ref().and_then(|user| {
    if user != &ssh_user {
        Some(format!("sudo -u {}", user))
    } else {
        None
    }
});
```

flake.nix
```diff
@@ -0,0 +1,76 @@
+# SPDX-FileCopyrightText: 2020 Serokell <https://serokell.io/>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+{
+  description = "A Simple multi-profile Nix-flake deploy tool.";
+
+  inputs = {
+    nixpkgs.url = "github:NixOS/nixpkgs/nixpkgs-unstable";
+    naersk = {
```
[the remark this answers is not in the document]
notgne2: I prefer naersk as it can avoid IFD or secondary lockfile generation, and tends to evaluate a lot faster

src/utils/deploy.rs
notgne2: `.as_deref()` instead, and some other tweaks, but it always seems to produce totally different assembly when I test it in isolated cases, and I don't imagine it's any more efficient so I think it's better to stick with `match` for this kind of thing

src/utils/mod.rs
notgne2: This again I think would turn out more confusing

src/main.rs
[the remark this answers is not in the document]
balsoft: Well, yes. More documentation will solve this issue.

src/utils/mod.rs
notgne2: Oh I think I get what you mean now, I didn't add any form of escaping at all, I probably should

src/main.rs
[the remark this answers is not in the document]
notgne2: Ah I think I missed what you meant, I've added a note about it now. The reason it will add all of them is because I can see a use case where only the order of some of the profiles would be important, the rest can run arbitrarily afterwards, though if limiting the profiles which are automatically deployed is desirable I can change this

src/utils/mod.rs
notgne2: Well, it looks like Nix doesn't let you have `#` in your path either, and has no escaping at all, so I guess there's no point fixing that if `nix` will fail when it encounters it anyway, and it would just create more inconsistencies

balsoft:
> I haven't tested bootstrap at all and I'm not really sure about the purpose so it makes sense that it might not all be logically sound. What exactly is the use case?

The use case is to run some commands when the profile is first created. E.g. in Ment, this is used to load the sql seed into the database.

notgne2:
@balsoft shouldn't this be done as a part of the activation script? Having an easy way to ensure it will only run upon profile creation might afford some convenience, though it's also unusual to have a deployment property which won't take effect if modified unless the profile were to be erased first. Eventually I'd like to add full and better support for stuff like this though, with automatic migrations and whatnot.

balsoft:
@notgne2 then it makes sense to just remove `bootstrap` completely.

.gitignore
```diff
@@ -0,0 +1,7 @@
+# SPDX-FileCopyrightText: 2020 Serokell <https://serokell.io/>
+#
+# SPDX-License-Identifier: MPL-2.0
+
+/target
+result*
+/examples/system/bare-system.qcow2
```
mkaito, line 7: newline

README.md
```diff
@@ -0,0 +1,180 @@
...
+## API
+
+### Profile
+
+This is the core of how `deploy-rs` was designed, any number of these can run on a node, as any user (see further down for specifying user information). If you want to mimick the behaviour of traditional tools like NixOps or Morph, try just defining one `profile` called `system`, as root, containing a nixosSystem, and you can even similarly use [home-manager](https://github.com/nix-community/home-manager) on any non-privileged user.
+
+```nix
+{
+  # A derivation containing your required software, and a script to activate it in `${path}/activate`
+  # For ease of use, `deploy-rs` provides a function to easy all this required activation script to any derivation
```
mkaito, line 37: Grammar has left the chat

README.md
```diff
@@ -0,0 +1,180 @@
...
+}
+```
+
+### Node
+
+This defines a single node/server, and the profiles you intend it to run.
+
+```nix
+{
+  # The hostname of your server, don't worry, this can be overridden at runtime if needed
```
mkaito, line 55:
```suggestion
  # The hostname of your server. Can be overridden at invocation time with a flag.
```

README.md
```diff
@@ -0,0 +1,180 @@
...
+  # This is the user that the profile will be deployed to (will use sudo if not the same as above).
+  # If `sshUser` is specified, this will be the default (though it will _not_ default to your own username)
+  user = "root";
+
+  # This is an optional list of arguments that will be passed to SSH.
+  sshOpts = [ "-p" "2121" ];
+
+  # Fast connection to the node. If this is true, copy the whole closure instead of letting the node substitute.
+  # This defaults to `false`
+  fastConnection = false;
```
mkaito, line 108: Fast connection or not, wouldn't substituting always be faster anyway?

README.md
```diff
@@ -0,0 +1,180 @@
...
+
+  # If the previous profile should be re-activated if activation fails.
+  # this defaults to `true`
+  autoRollback = true;
+
+  # If the node should wait for `deploy` to connect for a second time after activation, to confirm the server has not been ruined.
+  # This defaults to `false`, though it is strongly recommend you activate it if you value safety
```
mkaito, line 115: You make it sound like it would be downright stupid to not enable this, but you don't enable it by default. This seems contradictory. It would make more sense to enable it by default, and give a list of reasons you might want to disable it.

README.md
```diff
@@ -0,0 +1,180 @@
...
+  magicRollback = true;
+
+  # The path which deploy-rs will use for temporary files, this is currently only used by `magicRollback` to create an inotify watcher in
+  # If not specified, this will default to `/tmp/deploy-rs`
+  # (if `magicRollback` is in use, this _must_ be writable by `user`)
+  tempPath = "/home/someuser/.deploy-rs";
+}
+```
+
+### Putting it together
```
mkaito, line 125: Feels like this is an overview section with examples that should be near the top, with more detailed sections following it.

README.md
```diff
@@ -0,0 +1,180 @@
...
+  };
+
+  # This is highly advised, and will prevent many possible mistakes
+  checks = builtins.mapAttrs (system: deployLib: deployLib.deployChecks self.deploy) deploy-rs.lib;
+}
+```
+
+There are full working deploy-rs Nix expressions in the [examples folder](./examples), and there is a JSON schema [here](./interface.json) which is used internally by the `deployChecks` mentioned above to validate your expressions.
+
+## Idea
```
mkaito, line 159: Also a section that should almost certainly be near the top.

README.md
```diff
@@ -0,0 +1,180 @@
...
+
+`deploy-rs` is a simple Rust program that will take a Nix flake and use it to deploy any of your defined profiles to your nodes. This is _strongly_ based off of [serokell/deploy](https://github.com/serokell/deploy), designed to replace it and expand upon it.
+
+This type of design (as opposed to more traditional tools like NixOps or morph) allows for lesser-privileged deployments, and the ability to update different things independently of eachother.
+
+## Things to work on
```
mkaito, line 165: Why is this here instead of youtrack?

interface.json
```diff
@@ -0,0 +1,114 @@
...
+                                }
+                            ]
+                        }
+                    },
+                    "additionalProperties": false
+                }
+            }
+        }
+    ]
+}
```
mkaito, line 114: newline

interface.json.license
```diff
@@ -0,0 +1,3 @@
+SPDX-FileCopyrightText: 2020 Serokell <https://serokell.io/>
+
+SPDX-License-Identifier: MPL-2.0
```
mkaito, line 3: newline

mkaito (changes requested):
Do yourself a favour and configure your editor to add newlines at EOL.

flake.nix
```diff
@@ -0,0 +1,76 @@
...
+          };
+
+          # DEPRECATED
+          checkSchema = checks.schema;
+
+          deployChecks = deploy: builtins.mapAttrs (_: check: check deploy) checks;
+
+          checks = {
+            schema = deploy: pkgs.runCommandNoCC "jsonschema-deploy-system" { } ''
+              ${pkgs.python3.pkgs.jsonschema}/bin/jsonschema -i ${pkgs.writeText "deploy.json" (builtins.toJSON deploy)} ${./interface/deploy.json} && touch $out
```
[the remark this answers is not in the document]
notgne2: It's only used for checks so nothing really matters, but https://docs.rs/jsonschema/0.4.0/jsonschema/ exists and that could be nicer

flake.nix
```diff
@@ -0,0 +1,76 @@
...
+                  ${activate}
+                '';
+                executable = true;
+                destination = "/deploy-rs-activate";
+              })
+            ];
+          };
+
+          # DEPRECATED
+          checkSchema = checks.schema;
```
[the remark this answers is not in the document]
notgne2: There might be a few configurations using it that way still, it doesn't cost us anything to keep that alias for a while

README.md
notgne2: I do not remember writing this and I'm checking my carbon monoxide detectors tbh

README.md
notgne2: I believe this will cause the node to pull from it's own substituters too, so this flag should be set to `true` if deploying to for instance, a local machine in your home, where your LAN is almost definitely faster than the internet

README.md
notgne2: It's off by default as it's unexpected behaviour, I feel like people might try using this tool to change their SSH port for instance, then get fustrated when it refuses to let them. Maybe if I make it more clear that `deploy-rs` does this then it would be ok to have as a default

README.md
notgne2: Not instead of, in addition to, this is a public project so hence the documentation is not only for us, it can give other people an understanding of the direction of the project

mkaito: In the case of public open source projects (that we expect anyone else to actually care about), I believe our policy is to use github issues instead of youtrack.

flake.nix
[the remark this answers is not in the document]
mkaito: I say remove it, let them break, and then fix them.

balsoft:
Suggestion: before deploying, show what's going to happen (i.e. dump all the profiles fully evaluated) and ask user for confirmation.

flake.nix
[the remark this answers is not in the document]
notgne2: I've now addressed the only usage of it I can find (blend-app) in a PR so I'm going to remove it now

flake.nix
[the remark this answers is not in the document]
notgne2: Thanks to [this](https://github.com/Stranger6667/jsonschema-rs/blob/c95a8d590366cfec6496e518fbd57369e41271c0/.gitignore#L3) line of code, I will not be using it. I'll stick with the Python checker for now

notgne2:
@balsoft Do you just mean print a list of profiles (or node+profiles) if it automatically does multiple when one is not specified?

mkaito:
I think we're done here. We wanted to review existing code, and we did. Let's merge this into master and do any future work in feature branches.

mkaito (approved):

balsoft:
> I think we're done here. We wanted to review existing code, and we did. Let's merge this into master and do any future work in feature branches.

I agree.

> @balsoft Do you just mean print a list of profiles (or node+profiles) if it automatically does multiple when one is not specified?

I'll make an issue.

balsoft (approved):

mkaito:
I've merged into master

balsoft:
@mika https://github.com/serokell/deploy-rs/issues/4

mika:
@balsoft you pinged the wrong user

balsoft:
Whoops, sorry :) It's hard to remember everyone's usernames on all the platforms sometimes.

PR #2, opened 2020-11-14
[OPS-1085] Add --impure, remove --extra-build-args

mkaito:
* --extra-build-args never worked, removed
* Add --impure to pass along to nix flake

Tested by deploying Albali, which requires `--impure`.


balsoft:
Why is this needed? `deploy <...> -- --impure` should work.

mkaito:
Yeah, this is unnecessary

PR #5, opened 2020-11-15
Add activation presets for NixOS and noop activation

balsoft:
It turns out that some activation scripts are very common, and I think having special functions for them would be nice.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -64,6 +64,10 @@
             ];
           };
 
+          nixosActivate = base: setActivate base "$PROFILE/bin/switch-to-configuration switch";
+
+          noopActivate = base: setActivate base ":";
+
           deployChecks = deploy: builtins.mapAttrs (_: check: check deploy) {
             schema = deploy: pkgs.runCommandNoCC "jsonschema-deploy-system" { } ''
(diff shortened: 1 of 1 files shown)


notgne2:
I feel like these should be namespaced

balsoft:
Yeah, but then `setActivate` would look very strange unnamespaced. Move everything into `lib.activate.{noop,nixos,custom}` and `setActivate = builtins.trace "lib.setActivate is deprecated, use lib.activate.custom instead!" lib.activate.custom` ?

notgne2:
sounds good to me

notgne2:
Though I will add, I think the warning should actually refer to the existence of `noop` and `nixos` as options too, as like you said in the PR message, those are very common

flake.nix
```diff
@@ -48,25 +48,32 @@
...
+                  executable = true;
+                  destination = "/deploy-rs-activate";
+                })
+              ];
+            };
 
-          noopActivate = base: setActivate base ":";
+            nixos = base: setActivate base "$PROFILE/bin/switch-to-configuration switch";
+
+            noop = base: setActivate base ":";
```
notgne2, line 75: doesn't this mean any usage of `activate.{nixos, noop}` is going to produce a deprecation warning?

PR #6, opened 2020-11-19
Check auto-rollback flag before deactivating

mkaito:
I wrapped every call to `deactivate()` I could find, except one in `activation_confirmation()` that I'm not sure should be prevented.

### DIFF
--- a/src/activate.rs
+++ b/src/activate.rs
@@ -288,7 +288,9 @@ pub async fn activate(
     match nix_env_set_exit_status.code() {
         Some(0) => (),
         a => {
-            deactivate(&profile_path).await?;
+            if auto_rollback {
+                deactivate(&profile_path).await?;
+            }
             return Err(ActivateError::SetProfileExitError(a));
         }
     };
@@ -302,7 +304,9 @@ pub async fn activate(
     {
         Ok(x) => x,
         Err(e) => {
-            deactivate(&profile_path).await?;
+            if auto_rollback {
+                deactivate(&profile_path).await?;
+            }
             return Err(e);
         }
     };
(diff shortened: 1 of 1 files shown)


src/activate.rs
```diff
@@ -324,7 +330,9 @@ pub async fn activate(
         {
             Ok(()) => {}
             Err(err) => {
-                deactivate(&profile_path).await?;
+                if auto_rollback {
+                    deactivate(&profile_path).await?;
```
balsoft, line 334: I think this `deactivate` should actually run without `auto_rollback` because it's not an application failure but a magic rollback failure.

mkaito: I'm actually not sure whether we should roll back just because magic rollback failed to set up, but that's a different discussion I guess.

The point is that something failed and we're rolling back. I think that if you set auto-rollback to false, you would expect to, well, not roll back on failure.

notgne2: I definitely think it should, if you enable magic rollback you want the system to roll back if it cannot be confirmed, you do not want the safety of it to be undermined by any failure, and technically you can consider this failure to confirm too

mkaito: This is not where the magic rollback does its rollback. This is the rollback that happens if we fail to set up magic rollback.

mkaito: I've removed the conditional. Guess we'll find out via usage whether this is correct or not.

notgne2: I'm adamant having it always deactivate at that point is intended behaviour, the purpose of magic rollback is to provide certain garuantees, I trust it to bring my machine back if I break it, I don't want accidentally breaking some file permissions beforehand to bypass the whole system and let me ruin my server.

mkaito: Then maybe you want to key the canary by node and profile name, or at least figure out a way to ensure an unprivileged user doesn't fail to set up the canary just because he's not root, which is what I most often seem to run into.

PR #9, opened 2020-11-25
Fix running & building under macOS (Big Sur)

antifuchs:
I've been meaning to try deploy-rs, but I run macOS; since it doesn't currently build there (the `activate` binary has a linux-only inotify dependency, and there was a version incompatibility that prevented building on my OS release), here's a patch that makes it work.

I think this PR ought to retain compatibility with Linux.

## What's it do in detail?

* Bumps nixpkgs to a version that ships rust 1.47, as older binaries don't correctly link on macOS 11.0
* Special-cases out the `activate` binary on macOS (and activates `singleStep` mode there for https://github.com/nmattia/naersk/issues/127
* Adds some native build flags necessary to get a working build env on macOS.

Hope this looks reasonable! I'm still evaluating deploy-rs, but it does show `--help` on macOS with this change, at least (:

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -39,11 +39,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1601961544,
-        "narHash": "sha256-uuh9CkDWkXlXse8IcergqoIM5JffqfQDKsl1uHB7XJI=",
+        "lastModified": 1605716121,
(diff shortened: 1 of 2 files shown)


notgne2:
This looks good, in the future I'd maybe like to make the `activate` binary work on macOS too (either removing `magicRollback` support or using the equivilent APIs)

antifuchs:
I think the `notify` crate might be a match for this use case - it is cross-platform and might even have a nicer API than inotify does (:

Issue #13, opened 2020-11-25
Likely confusion between target system & system running deploy


antifuchs:
I finally got to the "activate" stage in my porting to deploy-rs (while running deploy on my macOS Big Sur machine), and unfortunately the deploy failed with:

```
 INFO  deploy > Evaluating flake in ./
 WARN  deploy > The following profiles are going to be deployed:
[monitor.system]
user = "root"
ssh_user = "asf"
path = "/nix/store/rscx6ya03wvd4qj2vv80pz1kazw7whrh-activatable-nixos-system-monitor-21.03.20201121.1ffd7cf"
hostname = "monitor.example.com"
ssh_opts = []

 INFO  deploy::utils::push > Building profile `system` for node `monitor`
warning: Git tree '/Users/asf/home' is dirty
 INFO  deploy::utils::deploy > Activating profile `system` for node `monitor`
sudo: /nix/store/n8hj08m4n6bwbvdm92ka59rjmaz29hvg-deploy-rs-0.1.0/bin/activate: command not found
Connection to monitor.example.com closed.
 ERROR deploy                > Failed to deploy profile: Activation over SSH resulted in a bad exit code: Some(1)
```

What's more, the closure that lands on the linux machine, printed above (`/nix/store/n8hj08m4n6bwbvdm92ka59rjmaz29hvg-deploy-rs-0.1.0/bin/`) contains a `deploy` binary that's built for macOS.

My guess is that the closure that gets copied to the target machine is compiled for the machine running the deploy. Maybe there's one more thing that needs building with the correct target platform set.

antifuchs:
oh, just in case it's unclear, I am using the `x86_64-linux` platform activate function:

``` nix
    deploy = {
      nodes = {
        "monitor" = {
          hostname = "monitor.example.com";

          profiles.system = {
            path = deploy-rs.lib.x86_64-linux.activate.nixos self.nixosConfigurations."monitor";
          };
        };
    };
```

balsoft:
Ah, yes, right. @notgne2 how about we choose and ship the `activate` binary in the `activate.custom` function instead of doing so at runtime?

antifuchs:
Traced this down into the guts of the deploy sources (https://github.com/serokell/deploy-rs/blob/3da8c6d048be03ecaf5caa54d945740ab88be465/src/utils/mod.rs#L165-L166), where we take the currently-running executable, which isn't guaranteed to work on the target.

balsoft:
[...]

PR #14, opened 2020-11-26
Use the stored activate binary appropriate for target platform

antifuchs:
This addresses #13.

This PR rips out the `current_exe` field, preferring to write an `activate-rs` script that points to the platform-appropriate activation binary.

Then, the deploy binary attempts to use that activate-rs script to activate the profile closure.

I've successfully run a deploy to an amd64 linux installation from my intel macOS machine, so I suspect this might work (:

### DIFF
--- a/src/utils/push.rs
+++ b/src/utils/push.rs
@@ -103,9 +103,7 @@ pub async fn push_profile(
             .arg("-k")
             .arg(local_key)
             .arg(&deploy_data.profile.profile_settings.path)
-            .arg(&super::deploy_path_to_activate_path_str(
-                &deploy_defs.current_exe,
-            )?)
+            .arg(&super::deploy_path_to_activate_path_str(&deploy_defs)?)
             .status()
             .await
             .map_err(PushProfileError::SignError)?;
@@ -150,9 +148,7 @@ pub async fn push_profile(
         .arg("--to")
         .arg(format!("ssh://{}@{}", deploy_defs.ssh_user, hostname))
         .arg(&deploy_data.profile.profile_settings.path)
-        .arg(&super::deploy_path_to_activate_path_str(
-            &deploy_defs.current_exe,
-        )?)
+        .arg(&super::deploy_path_to_activate_path_str(&deploy_defs)?)
         .env("NIX_SSHOPTS", ssh_opts_str)
         .status()
         .await
--- a/src/utils/push.rs
+++ b/src/utils/push.rs
@@ -148,7 +148,6 @@ pub async fn push_profile(
         .arg("--to")
         .arg(format!("ssh://{}@{}", deploy_defs.ssh_user, hostname))
         .arg(&deploy_data.profile.profile_settings.path)
-        .arg(&super::deploy_path_to_activate_path_str(&deploy_defs)?)
         .env("NIX_SSHOPTS", ssh_opts_str)
         .status()
         .await
--- a/src/utils/push.rs
+++ b/src/utils/push.rs
@@ -103,7 +103,6 @@ pub async fn push_profile(
             .arg("-k")
             .arg(local_key)
             .arg(&deploy_data.profile.profile_settings.path)
-            .arg(&super::deploy_path_to_activate_path_str(&deploy_defs)?)
             .status()
             .await
             .map_err(PushProfileError::SignError)?;
--- a/src/utils/push.rs
+++ b/src/utils/push.rs
@@ -4,6 +4,7 @@
 
 use std::process::Stdio;
 use tokio::process::Command;
+use std::path::Path;
 
 use thiserror::Error;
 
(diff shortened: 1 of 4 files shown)


balsoft:
There's a slight issue if someone tries to run a `deploy-rs` binary with this change on a flake that depends on a `deploy-rs` flake without this change. We probably need to handle this and when `activate-rs` is not found, throw an error that tells the user to run the same version as is pinned in the deployed flake.

balsoft:
This patch is fail-early and should make error messages a bit more clear:

```
diff --git a/src/utils/push.rs b/src/utils/push.rs
index 546ba3d..c6c28b9 100644
--- a/src/utils/push.rs
+++ b/src/utils/push.rs
@@ -4,6 +4,7 @@
 
 use std::process::Stdio;
 use tokio::process::Command;
+use std::path::Path;
 
 use thiserror::Error;
 
@@ -15,6 +16,12 @@ pub enum PushProfileError {
     BuildError(std::io::Error),
     #[error("Nix build command resulted in a bad exit code: {0:?}")]
     BuildExitError(Option<i32>),
+    #[error("Activation script deploy-rs-activate does not exist in profile.\n\
+             Did you forget to use deploy-rs#lib.<...>.activate.<...> on your profile path?")]
+    DeployRsActivateDoesntExist,
+    #[error("Activation script activate-rs does not exist in profile.\n\
+             Is there a mismatch in deploy-rs used in the flake you're deploying and deploy-rs command you're running?")]
+    ActivateRsDoesntExist,
     #[error("Failed to run Nix sign command: {0}")]
     SignError(std::io::Error),
     #[error("Nix sign command resulted in a bad exit code: {0:?}")]
@@ -91,6 +98,16 @@ pub async fn push_profile(
         a => return Err(PushProfileError::BuildExitError(a)),
     };
 
+    if ! Path::new(format!("{}/deploy-rs-activate", deploy_data.profile.profile_settings.path).as_str()).exists() {
+        return Err(PushProfileError::DeployRsActivateDoesntExist);
+    }
+
+    if ! Path::new(format!("{}/activate-rs", deploy_data.profile.profile_settings.path).as_str()).exists() {
+        return Err(PushProfileError::ActivateRsDoesntExist);
+    }
+
+
+
     if let Ok(local_key) = std::env::var("LOCAL_KEY") {
         info!(
             "Signing key present! Signing profile `{}` for node `{}`",

```

Although I guess it can be submitted in a separate PR.

antifuchs:
Making it fail early is a great solution; you have push permissions to my branch, but I can commit it also (:

balsoft:
Ah, ok, forgot about this feature of github :) I'll push this.

antifuchs:
One more thing that might make this less of a problem might be to find&document a good way of running the deploy via an `apps` entry in flake.nix itself: that way, the thing being run would be much more likely to be of the version that is pinned. I'll try to find a good one & submit a change to the readme next.

balsoft:
We're using either `defaultApp = deploy-rs.defaultApp` or `apps = builtins.mapAttrs (_: deploy: { inherit deploy; }) deploy-rs.defaultApp` in our repos.

antifuchs:
Oooh, the latter sounds really good - I like that it all turns into `nix run ./#deploy` when you set it up like that.

kirelagin:
_(annoying prescriptivist hated by everyone joins the chat)_

We are trying to keep this repo REUSE compliant and track copyright information correctly, so, @antifuchs, would you mind adding an `SPDX-FileCopyrightText` line in the header of the files that you made non-trivial changes to?

_We should probably add this to our contributing guidelines._

antifuchs:
Hah, that's totally legit, @kirelagin. I'll check with my employer's legal team (who are currently enjoying a long weekend) to ensure I'm doing it right - will update this PR on Monday.

kirelagin:
> I'll check with my employer's legal team

Sounds great, thanks for understanding!

Just to be clear, we do not require a CLA, assigning the copyright to us, or anything like that. We would just like to explicitly record for posterity that it is you who are the author (and thus the copyright holder) of what you authored :).

However, all this applies only if you are writing this code on your own time. If you are doing this as part of your job, then you would need to get a permission from your employer to release your works under the license that we use (MPL-2.0), and the copyright header should contain not your, but your employer’s name.

src/utils/push.rs
```diff
@@ -91,6 +98,16 @@ pub async fn push_profile(
         a => return Err(PushProfileError::BuildExitError(a)),
     };
 
+    if ! Path::new(format!("{}/deploy-rs-activate", deploy_data.profile.profile_settings.path).as_str()).exists() {
+        return Err(PushProfileError::DeployRsActivateDoesntExist);
+    }
+
+    if ! Path::new(format!("{}/activate-rs", deploy_data.profile.profile_settings.path).as_str()).exists() {
```
notgne2, line 105: This is good, but can we add this to one of our built-in flake checks too (like we do with `deploy-rs-activate`?)

notgne2 (changes requested):
This is all really good stuff, but there is a lot more code we can get rid of thanks to this, such as the `NotNixStored` error

src/utils/push.rs
balsoft: Ah, yes, forgot about it.

notgne2 (approved):

antifuchs:
Oh! There's a Checked with my legal friends at work & all is well, as expected - I added the copyright header & we're good to go.

Issue #27, opened 2020-12-12
Master doesn't build with the rust toolchain in 20.09 release


antifuchs:
I was surprised to find that the latest HEAD of deploy-rs doesn't build with rust 1.45.2, the version that's in the 20.09 release:

```
:;    cargo +1.45.2 test
   Compiling smol_str v0.1.17
   Compiling thin-dst v1.1.0
   Compiling rustc-hash v1.1.0
   Compiling text_unit v0.1.10
   Compiling cbitset v0.2.0
error[E0658]: `while` is not allowed in a `const fn`
  --> /Users/asf/.cargo/registry/src/github.com-1ecc6299db9ec823/smol_str-0.1.17/src/lib.rs:58:9
   |
58 | /         while i < text.len() {
59 | |             buf[i] = text.as_bytes()[i];
60 | |             i += 1
61 | |         }
   | |_________^
   |
   = note: see issue #52000 <https://github.com/rust-lang/rust/issues/52000> for more information

error: aborting due to previous error

For more information about this error, try `rustc --explain E0658`.
error: could not compile `smol_str`.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed
```

Looks like smol_str (pulled in by rowan, via rnix) is using const fn's in a slightly too-modern way.

PR #28, opened 2020-12-12
Debump the smol_str version to 0.1.16 to build on rust 1.45.2

antifuchs:
This addresses #27: Rust 1.45.2 is what ships in the 20.09 release, and smol_str 0.1.17 is
incompatible with it.

I think there's still a puzzle piece missing, namely some form of CI that asserts deploy-rs keeps working with the nix releases that you wish to support (I am making the assumption that 20.09 is a reasonable target for now).

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -27,6 +27,7 @@ thiserror = "1.0"
 toml = "0.5"
 yn = "0.1"
 rnix = "0.8"
+smol_str = "=0.1.16" # required by rnix, but 0.1.17 doesn't build on nixos-20.09
 
 
 [[bin]]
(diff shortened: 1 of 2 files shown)


Cargo.toml
```diff
@@ -27,6 +27,7 @@ thiserror = "1.0"
 toml = "0.5"
 yn = "0.1"
 rnix = "0.8"
+smol_str = "=0.1.16" # required by rnix, but 0.1.17 doesn't build on nixos-20.09
```
notgne2, line 30: I'd like the note to include the Rust version of NixOS 20.09 as well as the Rust version which is new enough to not have this issue

antifuchs: Good point, I'll add it.

PR #29, opened 2020-12-14
Build "activate" binary on all platforms

antifuchs:
I felt bad about the gross cargo workaround I left in the codebase for Darwin in #14, so I went in and ripped out the `inotify` dependency, in favor of a `notify` dependency - which is multiplatform and does a bunch of nice things for us (debounces events, automatically watches with less-sophisticated mechanisms if the watcher-of-the-day isn't working).

That required upgrading the tokio crate dep, but the upgrade was pretty painless.

The only thing I am not super happy with is the `eprintln` below; I don't think it's right to panic, as explained in the comment, but there's also no other way to tell the activation task that something is wrong.

What do you all think?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -2,18 +2,18 @@
 # It is not intended for manual editing.
 [[package]]
 name = "aho-corasick"
-version = "0.7.13"
+version = "0.7.15"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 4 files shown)


balsoft (approved):
Looks fine, but I'm not really a Rust expert.

notgne2:
I don't think the way we do panics or errors at this state really matters since the parent will probably be dead and there shouldn't even be an stdout available. These changes look good to me, and I've tested it as working.

PR #30, opened 2020-12-18
Fix fast_connection flag to enable substitute-on-destination correctly

notgne2:


src/utils/push.rs
```diff
@@ -134,7 +134,7 @@ pub async fn push_profile(
     let mut copy_command_ = Command::new("nix");
     let mut copy_command = copy_command_.arg("copy");
 
-    if let Some(true) = deploy_data.merged_settings.fast_connection {
+    if let None | Some(false) = deploy_data.merged_settings.fast_connection {
```
balsoft, line 137: Hm, why not
```suggestion
if deploy_data.merged_settings.fast_connection != Some(true) {
```

Am I missing something?

notgne2: That's much simpler, not sure why I decided to stick with destructuring/pattern matching

balsoft (approved):

Issue #32, opened 2020-12-19
Activation script runs forever


lovesegfault:
Even after the deploy has been successful and my `deploy .#myHost` command has exited, I can see the activation script running on my machine and pegging one CPU on 100% utilization.
```
❯ ps aux | rg -i deploy-rs
root     18497  100  0.0 953168   608 ?        Sl   20:57   0:41 /nix/store/5sdsrjrk314jv8zw8179bl65bkcg3dim-deploy-rs-0.1.0/bin/activate /nix/var/nix/profiles/system /nix/store/m8cv0h46jl4krqvwhf469561c3mlv1xs-activatable-nixos-system-foucault-21.03.20201218.9e67377 --temp-path /tmp --confirm-timeout 30 --magic-rollback --auto-rollback
bemeurer 18988  0.0  0.0   9700  4556 pts/2    R+   20:58   0:00 rg -i deploy-rs
```

This happens every single time I deploy to my machine.

notgne2:
Can you try running it with `--magic-rollback false` and see if this still happens?

lovesegfault:
Yeah, I can confirm the issue does _not_ happen if I disable magic rollback. And I can also reproduce it _every_ time I do a deploy _with_ magic rollback.

lovesegfault:
Actually, scratch that, seems like I just had it not happen. The deploy immediately after it, though, reproduced the problem again.

antifuchs:
@lovesegfault can you confirm that this happens on ba7f57086c41e5094f880e9bb7a117861bb4df5a as well as latest HEAD? I recently changed the activate script's magic rollback functionality, and while I see it work for me, I'm curious if that's what broke it for you...

kraem:
I'm having the same problem on 91e934fd9e1e06044f75a3e3ad998813c2a90fb5.

Having another issue with `magicRollback` where the `sshUser` is in group `wheel`. It asks for my sudo password like it should  but fails immediately after:
```
 INFO  deploy::utils::deploy > Success activating!
 INFO  deploy::utils::deploy > Attempting to confirm activation
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
 ERROR deploy                > Failed to deploy profile: Confirming activation over SSH resulted in a bad exit code (the server should roll back): Some(1)
``` 
Would this be a misconfiguration or a bug? I'm open to open another issue if this is the wrong forum..

antifuchs:
[...]

PR #36, opened 2020-12-21
Make magicRollback use pseudo-terminal allocation

kraem:
Solves issue with `magicRollback` where the `sshUser` is in group `wheel`. It asks for my sudo password like it should  but fails immediately after:
```
 INFO  deploy::utils::deploy > Success activating!
 INFO  deploy::utils::deploy > Attempting to confirm activation
sudo: a terminal is required to read the password; either use the -S option to read from standard input or configure an askpass helper
sudo: a password is required
 ERROR deploy                > Failed to deploy profile: Confirming activation over SSH resulted in a bad exit code (the server should roll back): Some(1)
```

### DIFF
--- a/src/main.rs
+++ b/src/main.rs
@@ -59,6 +59,9 @@ struct Opts {
     /// Override the SSH options used
     #[clap(long)]
     ssh_opts: Option<String>,
+    /// Use interactive TTY over SSH
+    #[clap(long)]
+    ssh_interactive_tty: bool,
     /// Override if the connecting to the target node should be considered fast
     #[clap(long)]
     fast_connection: Option<bool>,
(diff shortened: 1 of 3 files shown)


notgne2:
@kraem I don't see anything wrong with adding additional features, but as @lovesegfault said in #32, having that be the default can lead to less reproducibility and unexpected functionality. Can you put this behind a command-line feature argument for deploy-rs?

kraem:
Sure I can try, haven't written much rust before :)

kraem:
I'm not sure if this should be an `Option<bool>` instead. (and if this should be a per host configuration?)

notgne2:
I was about to revisit this and update it to work now there's some changes, but decided against it. I think `sshOpts` should be sufficient for this and can be used instead. Other than that, this PR would've been fine except for `&bool` which is unnecessary.

PR #39, opened 2020-12-25
Set /tmp as the CWD for activation

antifuchs:
Due to a nixos bug (https://github.com/NixOS/nixpkgs/issues/73404), the activation requires CWD to be writeable; one canonical writeable directory is /tmp, so I picked that. This fixes the issue for me, and I can install initrd secrets with the fix in place.

I guess it would be reasonable to let users choose their own, but I'm not sure how to properly do that - maybe via a custom activation script?

### DIFF
--- a/src/activate.rs
+++ b/src/activate.rs
@@ -138,7 +138,7 @@ pub async fn deactivate(profile_path: &str) -> Result<(), DeactivateError> {
 
     let re_activate_exit_status = Command::new(format!("{}/deploy-rs-activate", profile_path))
         .env("PROFILE", &profile_path)
-        .current_dir(&profile_path)
+        .current_dir("/tmp")
         .status()
         .await
         .map_err(DeactivateError::ReactivateError)?;
@@ -303,7 +303,7 @@ pub async fn activate(
 
     let activate_status = match Command::new(format!("{}/deploy-rs-activate", profile_path))
         .env("PROFILE", &profile_path)
-        .current_dir(&profile_path)
+        .current_dir("/tmp")
         .status()
         .await
         .map_err(ActivateError::RunActivateError)


notgne2:
I think this would be better done in the `activate.nixos` function in `flake.nix`.

We have `$PROFILE` so having PWD point to the profile isn't _needed_, but I might have still relied on that functionality in a few deployments (and some others may have too) and there's no point creating incompatibilities without a good reason

notgne2:
Also I'm against using `/tmp` without it being configurable, it's available in (most) NixOS configurations, though that's not all this tool should be able to work with

0x4A6F:
This issue can probably be closed, when NixOS/nixpkgs#73780 is merged.

antifuchs:
NixOS/nixpkgs#73780 won't make it into 20.09, right? I think we might need some method to customize the cwd anyway if not.

notgne2:
>I think we might need some method to customize the cwd anyway if not

Why not just change the NixOS activation helper to `nixos = base: custom base.config.system.build.toplevel "cd /tmp && $PROFILE/bin/switch-to-configuration switch";` in the `flake.nix`?

antifuchs:
Ah, I just found a case where your workaround is not quite optimal: When activation fails, the magic rollback will re-activate the previous generation, but if that's one that is built without this activation script, that generation can no longer properly be activated outside a writable directory.

szlend:
> Ah, I just found a case where your workaround is not quite optimal: When activation fails, the magic rollback will re-activate the previous generation, but if that's one that is built without this activation script, that generation can no longer properly be activated outside a writable directory.

Even if the bug was fixed in nixpkgs that would still be the case when you'd be deploying the fix for the first time. I would just re-deploy the previous generation with this workaround applied before doing any other changes if you want to safely rollback to it.

notgne2:
I'm closing this in favor of the now-merged #65

Issue #38, opened 2020-12-22
rollback even if I got "Success activating!"


bbigras:
I'm pretty sure I'm seeing a rollback even if I got "Success activating!".

Is there a log or something on the target to verify?

EDIT: I see 2 `hm-activate-bbigras` runs for the same deploy.

bbigras:
Here's a recording of it. I'm running it on my desktop (target is desktop too). The deploy is done at 01:38.

[![asciicast](https://asciinema.org/a/tDTskWPC83UERbFIUdS1F2Cnr.svg)](https://asciinema.org/a/tDTskWPC83UERbFIUdS1F2Cnr)

bbigras:
Here's a recording with some of the journalctl output:

[![asciicast](https://asciinema.org/a/cWyezNGsjm2mGg9nYu4AqrUu3.svg)](https://asciinema.org/a/cWyezNGsjm2mGg9nYu4AqrUu3)

kalbasit:
I'm having the same problem. A workaround is to deploy without the magic rollback:

```
deploy --magic-rollback false .#host
```

notgne2:
@kalbasit that helps narrow down the issue too, I'm assuming that the magic-rollback forked process is somehow unable to detect the "confirmation file" that gets produced to let that process know that the system is still functional and able to accept connections

notgne2:
I'm going to add an option which makes that process log to a file (it obviously can't log to stdout as it has to outlive the SSH connection), that way I can see what is going wrong. It's been configured to be extremely "cautious", if it encounters any error at all waiting for the confirmation file it will roll back (the preferable alternative to a bad config + a bad error leaving your machine potentially broken/inaccessible).

szlend:
I'm not sure if this is related but I see a lot of hanged activation scripts running on my server. Some of them are consuming 100% of CPU, like they're stuck on something.

bbigras:
> I'm not sure if this is related but I see a lot of hanged activation scripts running on my server. Some of them are consuming 100% of CPU, like they're stuck on something.

Did you see https://github.com/serokell/deploy-rs/issues/32 @szlend ?

antifuchs:
I'm seeing the same thing happen on my aarch64-linux machine. From what I can tell, one magic-rollback activation binary gets run; that terminates correctly and deploy success is announced. Then, *another* activation binary seems to get started (forked from somewhere?) which doesn't seem to be able to detect success & performs a rollback to the previous version.

[...]

PR #40, opened 2020-12-27
Replace logger, add more logging options including files

notgne2:
This replaces the logger which has both value in aesthetics, but also gives us file logging, which can be useful in debugging #38 and #32

EDIT: Updates to logging may have even _fixed_ #38 and #32


src/utils/mod.rs
```diff
@@ -19,6 +19,87 @@ macro_rules! good_panic {
...
+            log::Level::Error => "❌",
+            log::Level::Warn => "⚠️",
+            log::Level::Info => "ℹ️",
+            log::Level::Debug => "❓",
+            log::Level::Trace => "🖊️",
+        },
+        style(level, level.to_string()),
+        record.args()
+    )
+}
```
balsoft, line 64: Eww

notgne2: It won't take a closure :/

balsoft: I mean, at least pull out that `match`

notgne2: Done

mkaito: Please no

src/utils/mod.rs
```diff
@@ -19,6 +19,120 @@ macro_rules! good_panic {
...
+    record: &Record,
+) -> Result<(), std::io::Error> {
+    let level = record.level();
+
+    write!(
+        w,
+        "⭐ {0} {1} {0} {2}",
+        make_emoji(level),
+        style(level, level.to_string()),
+        record.args()
```
antifuchs, line 50: I don't know if flexi_logger supports fields or extended (per-process) metadata, but it would be good to get the PID of the process writing the log line here, to help me figure out https://github.com/serokell/deploy-rs/issues/38#issuecomment-753011354 (:

notgne2: That issue shouldn't exist anymore in this PR

balsoft:
Now magic rollback doesn't confirm that we're still able to authenticate via SSH, merely that the server is still reachable.

notgne2:
@balsoft I don't see how that's true, am I missing something here? The confirmation process works the same way, but instead of `deploy` waiting for the `activate` ssh connection to end before confirming, it waits for the new `wait` ssh connection to end (which occurs when the confirmation file is _created_), and instead of `activate` forking, it just ignores `SIGHUP`

balsoft:
I thought the second connection is started before the activation actually occurs, am I wrong?

notgne2:
- `ssh example.com activate &` (runs in the background, we don't care about it anymore)
- `ssh example.com wait` (waits for the confirmation file to exist, which happens after activation)
- `ssh example.com rm /tmp/confirmation-file` (actually confirm so the `activate` process from earlier doesn't rollback)

It _should_ be working like this, it only isn't if I have written it wrong or haven't thought about something correctly

balsoft:
Hm, ok, I've re-read the code and it seems that the logic is indeed 
correct.

notgne2:
I'll take this as a sign it's time for me to start documenting my code

notgne2:
(on second thought, there's a lot of code cleanup and documentation to be done everywhere, so I think I'll wait until post-merge)

mkaito (approved):
Seems like a solid improvement.

balsoft (approved):
Looks good, but haven't actually tested it.

Issue #31, opened 2020-12-19
Default boot entry is never updated


lovesegfault:
If I look at my system profiles I see this:
```
❯ ls /nix/var/nix/profiles
Permissions Size User Date Modified Name
lrwxrwxrwx    15 root 14 Dec 18:01  default -> default-87-link
lrwxrwxrwx    60 root 14 Dec  7:37  default-87-link -> /nix/store/nivqnkygpax10ph1s2f5m8km4nr8sc24-user-environment
drwxr-xr-x     - root  4 Jun  1:27  per-user/
lrwxrwxrwx    15 root 18 Dec 18:12  system -> system-434-link
lrwxrwxrwx   100 root 18 Dec 17:28  system-432-link -> /nix/store/nzd9if3g1m1z2b37694pkq3472mdrhnx-activatable-nixos-system-foucault-21.03.20201218.9e67377
lrwxrwxrwx   100 root 18 Dec 17:46  system-433-link -> /nix/store/yycw0bmzpi7l5bx5zkpk4nb79sfydyki-activatable-nixos-system-foucault-21.03.20201218.9e67377
lrwxrwxrwx   100 root 18 Dec 18:12  system-434-link -> /nix/store/m8cv0h46jl4krqvwhf469561c3mlv1xs-activatable-nixos-system-foucault-21.03.20201218.9e67377
```

This looks exactly right, my last system profile, 434, was a successful deploy and so `/nix/var/nix/profiles/system` points to it.

If, now, I go look at my bootloader (systemd-boot) config, I see this:
```
❯ cat /boot/loader/loader.conf
───────┬──────────────────────────────────────────────────
       │ File: /boot/loader/loader.conf
───────┼──────────────────────────────────────────────────
   1   │ timeout 2
   2   │ default nixos-generation-426.conf
   3   │ console-mode keep
───────┴──────────────────────────────────────────────────
```

And this is despite the fact that the correct generation _is_ available in the boot partition 
```
❯ ls /boot/loader/entries
Permissions Size User Date Modified Name
.rwxr-xr-x   748 root 18 Dec 18:12  nixos-generation-432.conf*
.rwxr-xr-x   748 root 18 Dec 18:12  nixos-generation-433.conf*
.rwxr-xr-x   748 root 18 Dec 18:12  nixos-generation-434.conf*
```

The current default happened right around the same time I switched to deploy-rs. Could it be that it's just not updating that parameter?

lovesegfault:
[...]

PR #42, opened 2021-01-05
Remove default loader entry on NixOS activation

notgne2:
This should work around #31 in a backwards-compatible manner, it might be better to update that line rather than delete it (though the default without this line is the latest generation, which should be fine), but that would be more complex and we may not need it.

We should of course still fix this upstream in nixpkgs to eventually remove the need for this workaround

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -80,7 +80,12 @@
               ];
             };
 
-            nixos = base: custom base.config.system.build.toplevel "$PROFILE/bin/switch-to-configuration switch";
+            nixos = base: custom base.config.system.build.toplevel ''
+              $PROFILE/bin/switch-to-configuration switch
+
+              # https://github.com/serokell/deploy-rs/issues/31
+              sed -i '/^default /d' /boot/loader/loader.conf
+            '';
 
             noop = base: custom base ":";
           };
--- a/flake.nix
+++ b/flake.nix
@@ -84,7 +84,7 @@
               $PROFILE/bin/switch-to-configuration switch
 
               # https://github.com/serokell/deploy-rs/issues/31
-              sed -i '/^default /d' /boot/loader/loader.conf
+              sed -i '/^default /d' ${base.config.boot.loader.efi.efiSysMountPoint}/loader/loader.conf || :
             '';
 
             noop = base: custom base ":";
--- a/flake.nix
+++ b/flake.nix
@@ -84,7 +84,9 @@
               $PROFILE/bin/switch-to-configuration switch
 
               # https://github.com/serokell/deploy-rs/issues/31
-              sed -i '/^default /d' ${base.config.boot.loader.efi.efiSysMountPoint}/loader/loader.conf || :
+              ${with base.config.boot.loader;
+              pkgs.lib.optionalString systemd-boot.enable
+              "sed -i '/^default /d' ${efi.efiSysMountPoint}/loader/loader.conf"}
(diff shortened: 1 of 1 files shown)


balsoft:
Wouldn't this mean deployment fails for non-systemd-boot systems? A lot of people still run GRUB.

balsoft:
Also, EFI mount point is not guaranteed to be `/boot`, it's `config.boot.loader.efi.efiSysMountPoint`

flake.nix
```diff
@@ -80,7 +80,12 @@
               ];
             };
 
-            nixos = base: custom base.config.system.build.toplevel "$PROFILE/bin/switch-to-configuration switch";
+            nixos = base: custom base.config.system.build.toplevel ''
+              $PROFILE/bin/switch-to-configuration switch
+
+              # https://github.com/serokell/deploy-rs/issues/31
+              sed -i '/^default /d' /boot/loader/loader.conf
```
balsoft, line 87:
```suggestion
              sed -i '/^default /d' ${config.boot.loader.efi.efiSysMountPoint}/loader/loader.conf
```

notgne2: oh, I forgot about that

notgne2:
@balsoft both of those issues should be fixed now, I'll test it on some GRUB and gummiboot systems

flake.nix
```diff
@@ -80,7 +80,12 @@
               ];
             };
 
-            nixos = base: custom base.config.system.build.toplevel "$PROFILE/bin/switch-to-configuration switch";
+            nixos = base: custom base.config.system.build.toplevel ''
+              $PROFILE/bin/switch-to-configuration switch
+
+              # https://github.com/serokell/deploy-rs/issues/31
+              sed -i '/^default /d' ${base.config.boot.loader.efi.efiSysMountPoint}/loader/loader.conf || :
```
szlend, line 87: I would check if the EFI loader is enabled, otherwise every non systemd-boot host will get `sed: can't read /boot/loader/loader.conf: No such file or directory` in their deploy logs.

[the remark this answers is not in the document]
notgne2: I'm not sure of that actually, I haven't tested it but I know that GRUB boots the newest `deploy-rs`-deployed profile at least _sometimes_ (when no old generations have been garbage collected), so it would be weird if it acts like systemd-boot only sometimes

notgne2: We could also just pipe stderr/stdout to /dev/null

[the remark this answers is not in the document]
balsoft: Ah, ok, you're right. GRUB does the right thing: https://github.com/NixOS/nixpkgs/blob/master/nixos/modules/system/boot/loader/grub/install-grub.pl#L491

It doesn't check whether `/nix/var/nix/profiles` contains `$defaultConfig`

mkaito: I'd rather see the unnecessary spam than accidentally suppress a useful error.

balsoft: So, I think actually checking if `config.boot.loader.systemd-boot.enable == true` is the best solution here. Something like

```suggestion
              ${with base.config.boot.loader;
              lib.optionalAttrs systemd-boot.enable
              "sed -i '/^default /d' ${efi.efiSysMountPoint}/loader/loader.conf"}
```

notgne2: I've added that change now

balsoft (approved):

PR #45, opened 2021-01-09
Code cleanup

notgne2:
Here is a bunch of clippy-motivated cleanups I've been meaning to make in the past, there's 3 more warnings left but those are more "justified" (they are warning of actual issues I will need to fix).

None of these changes should affect functionality (though until we have a proper tests system, we should manually test this)


zhenyavinogradov (approved):
Everything looks good

PR #48, opened 2021-01-14
Raise wait timeout

notgne2:
This stops problems from occuring when activation takes a long time to complete


balsoft (approved):
Maybe we should make this configurable at some point in the future.

PR #50, opened 2021-01-18
Update all flake inputs to latest versions

balsoft:
We have repinned nixpkgs, so it's a good idea to update it here. Also updating all the other inputs while we're at it.


notgne2 (approved):
We actually don't use the Serokell nixpkgs pin here, but updating regardless is still good

PR #59, opened 2021-01-29
Wait for activation

### LINKED ISSUE #49: Don't quit until SSH connections are closed
The current behaviour allows `deploy` to quit due to errors while leaving SSH connections open, printing messages while the program is closed and you have returned to your shell. The excess of logs is nice, though it is not very good looking, we should hold `deploy` open until these connections no longer exist

### LINKED ISSUE #58: deploy-rs gets stuck after deployment failure when magicRollback is enabled
For some reason, when `magicRollback` in enabled and the deployment script fails, deploy-rs hangs until the default timeout elapses instead of exiting immediately. Example of logs:
```
🚀 ℹ️ [deploy] [INFO] Activating profile `myprofile` for node `test`
🚀 ℹ️ [deploy] [INFO] Creating activation waiter
👀 ℹ️ [wait] [INFO] Waiting for confirmation event...
⭐ ❌ [activate] [ERROR] The activation script resulted in a bad exit code: Some(1)
<here it gets stuck until the timeout>
👀 ❌ [wait] [ERROR] Error waiting for activation: Timeout elapsed for confirmation
🚀 ❌ [deploy] [ERROR] Failed to deploy profile: Waiting over SSH resulted in a bad exit code: Some(1)
```

Disabling `magicRollback` makes it fail immediately as expected:
```
⭐ ❌ [activate] [ERROR] The activation script resulted in a bad exit code: Some(1)
🚀 ❌ [deploy] [ERROR] Failed to deploy profile: Activating over SSH resulted in a bad exit code: Some(1)
```

notgne2:
We previously started and abandoned the `ssh_activate` SSH session, which would let it spill into our shells, and mean that `wait` always has to time out (`deploy` will never error due to `activate` encountering an error). This PR will let the `activate` session end before the `wait` session, which occurs if an activation error is encountered, it then handles the error (if any) correctly.

fixes #58 and fixes #49 (I think)

### DIFF
--- a/src/deploy.rs
+++ b/src/deploy.rs
@@ -246,20 +246,25 @@ pub async fn deploy_profile(
             ssh_wait_command.arg(ssh_opt);
         }
 
-        tokio::pin! {
-            let ssh_wait_future = ssh_wait_command.arg(self_wait_command).status();
-            let ssh_activate_future = ssh_activate.wait_with_output();
-        }
+        let (send_activate, recv_activate) = tokio::sync::oneshot::channel();
+        let (send_activated, recv_activated) = tokio::sync::oneshot::channel();
+
+        tokio::spawn(async move {
+            send_activate
+                .send(ssh_activate.wait_with_output().await)
+                .unwrap();
+            send_activated.send(()).unwrap();
+        });
 
         tokio::select! {
-            x = ssh_wait_future => {
+            x = ssh_wait_command.arg(self_wait_command).status() => {
                 match x.map_err(DeployProfileError::SSHWaitError)?.code() {
                     Some(0) => (),
                     a => return Err(DeployProfileError::SSHWaitExitError(a)),
                 };
             },
-            x = ssh_activate_future => {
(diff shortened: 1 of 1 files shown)


notgne2:
Wait, no, this one isn't done

notgne2:
Not sure that last commit is implemented very well, and not sure what set of scenarios could lead to it being required either, but I think it's the right thing to add. Next is fixing early returns due to errors.

notgne2:
There, that doubles down on my bad ideas and makes this a fully functional solution. Unfortunately while I believe it _does_ work I don't think it should, I think it should be panicing randomly, so lets clean that up first.

notgne2:
This is also a pain to develop and debug with the activation binary being in such a sad state, that really needs cleaning up to centralize and solidify the error handling and rollbacks, and logging (especially in the case of logging errors that cause rollbacks) need to be fixed. But that is all for another PR

notgne2:
Ok, this is better now

balsoft (approved):
Maybe we should choose one name for target/node/host/server and keep using it everywhere

PR #62, opened 2021-02-04
Fix. Remove ssh:// prefix from parametr ssh_activate and ssh_confirm …

Zhevlovd:
Activate is trying connect with wrong user name. Build and "nix copy" steps are running fine. But Activate step is trying to connect with username "ssh".
Removing prefix ssh:// from ssh parameter fix it.


notgne2:
Closing now that #64 has been merged

PR #63, opened 2021-02-05
Evaluate deploy output lazily

balsoft:
Currently, we evaluate the `#deploy` output strictly. This means
- Longer eval times
- Extraneous evaluation errors with `--skip-checks`
- `-- --impure` even when the path we're currently deploying is pure
- etc.

With this change, evaluation happens lazily -- we only evaluate the nodes
and profiles we really need. It is only implemented for flaky Nix, and
it is on by default. To get the old behavior, one can specify
`--strict-eval`.

I have tested that this indeed dramatically increases evaluation speed
in all of our repos, and removes the need to deploy Agora with
`--impure`. Hooray!

## Notes for self if I were code reviewing this

- `r#"Foo "bar" baz"#` is a raw string evaluating to literally `Foo "bar" baz`
- `{{`/`}}` is a way to escape `{`/`}` in `std::fmt` strings

### DIFF
--- a/src/bin/deploy.rs
+++ b/src/bin/deploy.rs
@@ -52,10 +52,6 @@ struct Opts {
     #[clap(short, long)]
     skip_checks: bool,
 
-    /// Always evaluate deploy attribute strictly
-    #[clap(long)]
-    strict_eval: bool,
-
     /// Override the SSH user with the given value
     #[clap(long)]
     ssh_user: Option<String>,
@@ -160,12 +156,13 @@ enum GetDeploymentDataError {
     DecodeUtf8(#[from] std::string::FromUtf8Error),
     #[error("Error decoding the JSON from evaluation: {0}")]
     DecodeJson(#[from] serde_json::error::Error),
+    #[error("Impossible happened: profile is set but node is not")]
+    ProfileNoNode,
 }
 
 /// Evaluates the Nix in the given `repo` and return the processed Data from it
 async fn get_deployment_data(
     supports_flakes: bool,
-    strict_eval: bool,
     flake: &deploy::DeployFlake<'_>,
     extra_build_args: &[String],
 ) -> Result<deploy::data::Data, GetDeploymentDataError> {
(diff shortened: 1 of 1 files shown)


src/bin/deploy.rs
```diff
@@ -161,38 +165,83 @@ enum GetDeploymentDataError {
...
-        false => {
-            c
-                .arg("--strict")
-                .arg("--read-write-mode")
-                .arg("--json")
-                .arg("--eval")
-                .arg("-E")
-                .arg(format!("let r = import {}/.; in if builtins.isFunction r then (r {{}}).deploy else r.deploy", repo))
+            .arg(format!("{}#deploy", flake.repo));
+        match (&flake.node, strict_eval) {
```
notgne2, line 185: If there's a clean way to throw `flake.profile` into this match statement too, instead of having it as a child, then I would much prefer that.

notgne2 (commented):
This is really amazing to have overall, though for some reason on my own machine I wasn't able to save more than 1 out of 10 seconds, or 2 out of 20 seconds total runtime.

You mentioned that this would save us needing `--impure` for certain deployments, which is really neat too, but what is the reason for this? If it is because Nix is evaluating things outside of `deploy` needlessly, then I think some code is missing from this PR, as if you don't specify a node (such as just running `deploy .`) then wouldn't it still have issues even though it doesn't need to?

balsoft:
> You mentioned that this would save us needing --impure for certain deployments, which is really neat too, but what is the reason for this?

Case in point: `tezos-infra`. Tezos nodes require impure (because they depend on tezos stuff from tezos-packaging), but Agora nodes don't. Before this change, we would need `--impure` even when deploying Agora nodes with `--skip-checks`, which is rather annoying.

balsoft:
> for some reason on my own machine I wasn't able to save more than 1 out of 10 seconds, or 2 out of 20 seconds total runtime.

What is the `installable` (`repo#node.profile`) you're deploying?

balsoft:
For comparison, 
```
$ # Make sure that pegasus-infra is cached and all the dependencies are fetched (so that Nix only has to evaluate)
$ nix flake check github:serokell/pegasus-infra
$ # Before this change
$ RUST_LOG=trace deploy github:serokell/pegasus-infra#enif.system --skip-checks 2>&1 | ts -s '%.S'
00.000008 🚀 ❓ [deploy] [DEBUG] Checking for flake support
00.000061 🚀 ℹ️ [deploy] [INFO] Evaluating flake in github:serokell/pegasus-infra
00.001257 🚀 🖊️ [deploy] [TRACE] registering event source with poller: token=Token(1), interests=READABLE | WRITABLE
08.980935 🚀 🖊️ [deploy] [TRACE] deregistering event source from poller
08.981079 🚀 ℹ️ [deploy] [INFO] The following profiles are going to be deployed:
08.981091 [enif.system]
08.981098 user = "root"
08.981105 ssh_user = "balsoft"
08.981111 path = "/nix/store/8dgi8cps9sxihryq212m8y1zwr6hm8bg-activatable-nixos-system-enif-20.09.20201120.ee80272"
08.981118 hostname = "enif.pegasus.serokell.team"
08.981125 ssh_opts = ["-p", "17788"]
^C
$ # With this change
$ RUST_LOG=trace ./target/debug/deploy github:serokell/pegasus-infra#enif.system --skip-checks 2>&1 | ts -s '%.S'
00.000008 🚀 ❓ [deploy] [DEBUG] Checking for flake support
00.009012 🚀 ℹ️ [deploy] [INFO] Evaluating flake in github:serokell/pegasus-infra
00.010636 🚀 🖊️ [deploy] [TRACE] registering event source with poller: token=Token(1), interests=READABLE | WRITABLE
03.380132 🚀 🖊️ [deploy] [TRACE] deregistering event source from poller
03.380509 🚀 ℹ️ [deploy] [INFO] The following profiles are going to be deployed:
03.380522 [enif.system]
03.380529 user = "root"
03.380536 ssh_user = "balsoft"
03.380543 path = "/nix/store/8dgi8cps9sxihryq212m8y1zwr6hm8bg-activatable-nixos-system-enif-20.09.20201120.ee80272"
03.380550 hostname = "enif.pegasus.serokell.team"
03.380557 ssh_opts = ["-p", "17788"]
^C
```

src/bin/deploy.rs
```diff
@@ -161,38 +165,83 @@ enum GetDeploymentDataError {
...
+            (Some(node), false) => {
+                // We use --apply instead of --expr so that we don't have to deal with builtins.getFlake
+                c.arg("--apply");
+                info!("Evaluating the flake lazily; Use --strict-eval for old behavior");
```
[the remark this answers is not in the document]
balsoft: Removed `--strict-eval`

src/bin/deploy.rs
```diff
@@ -161,38 +165,83 @@ enum GetDeploymentDataError {
...
+                              }})
+                            "#,
+                            node, profile
+                        ))
+                    }
+                }
+            }
+            (_, _) => {
+                // We need to evaluate all profiles of all nodes anyway, so just do it strictly
+                &c
```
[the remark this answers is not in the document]
balsoft: Replaced with `--apply` thing too for consistency (and refactoring to a single `match`)

src/bin/deploy.rs
balsoft: Refactored into a single `match`

src/bin/deploy.rs
```diff
@@ -161,38 +165,83 @@ enum GetDeploymentDataError {
...
-    info!("Evaluating flake in {}", repo);
+    info!("Evaluating flake in {}", flake.repo);
 
-    let mut c = match supports_flakes {
-        true => Command::new("nix"),
-        false => Command::new("nix-instantiate"),
+    let mut c = if supports_flakes {
+        Command::new("nix")
+    } else {
+        Command::new("nix-instantiate")
```
notgne2, line 174: stupid nitpick here: I preferred this the way it was before

notgne2:
> > You mentioned that this would save us needing --impure for certain deployments, which is really neat too, but what is the reason for this?
> 
> Case in point: `tezos-infra`. Tezos nodes require impure (because they depend on tezos stuff from tezos-packaging), but Agora nodes don't. Before this change, we would need `--impure` even when deploying Agora nodes with `--skip-checks`, which is rather annoying.

Oh, so this is only relevant if there are some nodes that require and some nodes that don't witA?hin the same repo

> > for some reason on my own machine I wasn't able to save more than 1 out of 10 seconds, or 2 out of 20 seconds total runtime.
> 
> What is the `installable` (`repo#node.profile`) you're deploying?

Only 1 system in this repo, and first was blank, meaning both `system` and `home-manager` were being activated, second was just `system`. I'm not sure why there was an improvement for the first run (of both) at all.

I'll definitely try doing some more testing with more accurate measures too.

notgne2 (approved):

PR #70, opened 2021-02-27
Fix examples/system/README.md: `deploy` -> `deploy-rs`

talw:
I assume that the intention is to use `deploy-rs` and not its predecessor, or have I misunderstood?


notgne2 (approved):
Thanks!

Issue #46, opened 2021-01-09
Parallel Deploys


lovesegfault:
Would you consider allowing deploy-rs to deploy to multiple nodes simultaneously?

I wanted to attempt implementing that, but figured I'd check before so as not to end up wasting time.

notgne2:
Absolutely, but one thing to bare in mind is I would one day like to support deploying multiple profiles to the same node, which will require doing away with our `profilesOrder` attribute in favor of a dependency map.

I'm not suggesting you go through that hard work too (though I would obviously appreciate it), but if it ends up being relevant, if you could make your logic not too hard to develop into supporting paralell profile deployment that would make it easier for me in the future.

nrdxp:
I've been playing with tokio threads lately so if the author would still like to start on this or has already please share and I'll try to help :smile:

This would really be an infinitely scalable solution for large deployments.

jzbor:
I think a nice idea would be building in parallel (as this is the most work-intensive part) and deploying in serial to ensure low down times and some degree of determinism.

PR #80, opened 2021-03-30
Deploying multiple nodes with one command

### LINKED ISSUE #60: Command-line syntax for deploying multiple specific node/profile paths
This can be a bundle of features

-[ ] Allow to specify one or more nodes to deploy, f.e. `--node staging --node production`
-[ ] Allow a node to be excluded from the default if no flags are set, f.e. `defaultDeploy = false;` so that it has to be deployed with an explicit --node

ysndr:
**Draft PR**

- [x] implements running multiple deployments in sequence (resolves #60)
- [x] revokes previously succeeded deploys if a later one fails
- [ ] ~parallel deployment (#46)~

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,6 @@
 <!--
 SPDX-FileCopyrightText: 2020 Serokell <https://serokell.io/>
+SPDX-FileCopyrightText: 2021 Yannik Sander <contact@ysndr.de>
 
 SPDX-License-Identifier: MPL-2.0
 -->
@@ -16,18 +17,26 @@ Questions? Need help? Join us on Matrix: [`#deploy-rs:matrix.org`](https://matri
 
 Basic usage: `deploy [options] <flake>`.
 
-The given flake can be just a source `my-flake`, or optionally specify the node to deploy `my-flake#my-node`, or specify a profile too `my-flake#my-node.my-profile`. If your profile or node name has a `.` in it, simply wrap it in quotes, and the flake path in quotes (to avoid shell escaping), for example `'my-flake."myserver.com".system'`.
+Using this method all profiles specified in the given `<flake>` will be deployed (taking into account the [`profilesOrder`](#node)).
+
+ Optionally the flake can be constrained to deploy just a single node (`my-flake#my-node`) or a profile (`my-flake#my-node.my-profile`).
+
+If your profile or node name has a . in it, simply wrap it in quotes, and the flake path in quotes (to avoid shell escaping), for example 'my-flake."myserver.com".system'.
+
+Any "extra" arguments will be passed into the Nix calls, so for instance to deploy an impure profile, you may use `deploy . -- --impure` (note the explicit flake path is necessary for doing this).
 
 You can try out this tool easily with `nix run`:
 - `nix run github:serokell/deploy-rs your-flake`
 
-Any "extra" arguments will be passed into the Nix calls, so for instance to deploy an impure profile, you may use `deploy . -- --impure` (note the explicit flake path is necessary for doing this).
+In you want to deploy multiple flakes or a subset of profiles with one invocation, instead of calling `deploy <flake>` you can issue `deploy --targets <flake> [<flake> ...]` where `<flake>` is supposed to take the same format as discussed before.
+
+Running in this mode, if any of the deploys fails, the deploy will be aborted and all successful deploys rolled back. `--rollback-succeeded false` can be used to override this behavior, otherwise the `auto-rollback` argument takes precedent.
 
 If you require a signing key to push closures to your server, specify the path to it in the `LOCAL_KEY` environment variable.
 
 Check out `deploy --help` for CLI flags! Remember to check there before making one-time changes to things like hostnames.
 
-There is also an `activate` binary though this should be ignored, it is only used internally and for testing/hacking purposes.
+There is also an `activate` binary though this should be ignored, it is only used internally (on the deployed system) and for testing/hacking purposes.
 
 ## Ideas
 
@@ -79,7 +88,7 @@ A basic example of a flake that works with `deploy-rs` and deploys a simple NixO
 
 ### Profile
 
-This is the core of how `deploy-rs` was designed, any number of these can run on a node, as any user (see further down for specifying user information). If you want to mimick the behaviour of traditional tools like NixOps or Morph, try just defining one `profile` called `system`, as root, containing a nixosSystem, and you can even similarly use [home-manager](https://github.com/nix-community/home-manager) on any non-privileged user.
+This is the core of how `deploy-rs` was designed, any number of these can run on a node, as any user (see further down for specifying user information). If you want to mimic the behaviour of traditional tools like NixOps or Morph, try just defining one `profile` called `system`, as root, containing a nixosSystem, and you can even similarly use [home-manager](https://github.com/nix-community/home-manager) on any non-privileged user.
 
 ```nix
 {
@@ -128,7 +137,7 @@ This is the top level attribute containing all of the options for this tool
 {
   nodes = {
     # Definition format shown above
-    my-node = {}; 
+    my-node = {};
     another-node = {};
   };
 
--- a/src/bin/deploy.rs
+++ b/src/bin/deploy.rs
@@ -1,4 +1,5 @@
 // SPDX-FileCopyrightText: 2020 Serokell <https://serokell.io/>
+// SPDX-FileCopyrightText: 2021 Yannik Sander <contact@ysndr.de>
 //
 // SPDX-License-Identifier: MPL-2.0
 
@@ -7,6 +8,8 @@ use std::io::{stdin, stdout, Write};
 
 use clap::Clap;
 
+use deploy::{DeployFlake, ParseFlakeError};
+use futures_util::stream::{StreamExt, TryStreamExt};
 use log::{debug, error, info, warn};
 use serde::Serialize;
 use std::process::Stdio;
@@ -14,12 +17,16 @@ use thiserror::Error;
 use tokio::process::Command;
 
 /// Simple Rust rewrite of a simple Nix Flake deployment tool
-#[derive(Clap, Debug)]
+#[derive(Clap, Debug, Clone)]
 #[clap(version = "1.0", author = "Serokell <https://serokell.io/>")]
 struct Opts {
     /// The flake to deploy
-    #[clap(default_value = ".")]
-    flake: String,
+    #[clap(group = "deploy")]
+    target: Option<String>,
+
+    /// A list of flakes to deploy alternatively
+    #[clap(long, group = "deploy")]
+    targets: Option<Vec<String>>,
     /// Check signatures when using `nix copy`
     #[clap(short, long)]
     checksigs: bool,
@@ -77,6 +84,9 @@ struct Opts {
     /// Show what will be activated on the machines
     #[clap(long)]
     dry_activate: bool,
+    /// Revoke all previously succeeded deploys when deploying multiple profiles
+    #[clap(long)]
+    rollback_succeeded: Option<bool>,
 }
 
 /// Returns if the available Nix installation supports flakes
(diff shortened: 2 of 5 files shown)


notgne2:
Given the precedent deploy-rs has been setting in regards to safety, I would be inclined to prefer the extreme approach of handling a failure by rolling back every server that has already succeeded, then exiting. If that's too hard to implement, simply exiting without rolling back any previous nodes would be fine for now, but I think a CLI flag to disable this functionality and continue deploying other servers would be very useful.

ysndr:
> And one thing to bare in mind when working on the CLI arguments is the often overlooked `extra_build_args`, allowing me to do things like `deploy . -- --impure`, to instruct Nix to allow impurities despite deploy-rs having no `--impure` flag handling of it's own, so please make sure that remains working

I just added it as a new flag that takes several attributes using the -- syntax that shouldn't interfere with passing extra arguments

ysndr:
> rolling back every server that has already succeeded, then exiting.

We are thinking about what it means to rollback successful deploys.
Currently, when multiple profiles are deployed to a node its only the most recent (failed) profile that is rolled back if I see see that correctly.
Is this behaviour still preferable or should a node roll back all other profiles deployed as well?
I assume the latter but would like to check nevertheless.
Also should the behavior of the current implementation change to that as well (given --auto-rollback)? That is, if one profile fails all profiles are rolled back.

notgne2:
> > rolling back every server that has already succeeded, then exiting.
> 
> We are thinking about what it means to rollback successful deploys.
> Currently, when multiple profiles are deployed to a node its only the most recent (failed) profile that is rolled back if I see see that correctly.
> Is this behaviour still preferable or should a node roll back all other profiles deployed as well?
> I assume the latter but would like to check nevertheless.

I would like it to roll back all the profiles, and a flag and/or config option to avoid doing so, though now that I think about it a little more, that _does_ sound like a pain to implement, and so does rolling back old nodes, so feel free to skip implementing this at all.

> Also should the behavior of the current implementation change to that as well (given --auto-rollback)? That is, if one profile fails all profiles are rolled back.

If you do implement the above, then yes, I want this to take effect even when only deploying a single node.

---

One additional implementation concern I have is taking multiple `DeployFlake`s, while I guess it makes the CLI syntax simpler, it seems very unusual if it allows different `repo`s to be specified (as in, multiple `flake.nix`s being evaluated), but on the other hand, that might be nice too (i.e. if I have a different `flake.nix` for my local and online clusters, I can tell deploy-rs to deploy both in a single invocation), I just haven't decided how I feel about it yet.

ysndr:
> I would like it to roll back all the profiles, and a flag and/or config option to avoid doing so, though now that I think about it a little more, that does sound like a pain to implement, and so does rolling back old nodes, so feel free to skip implementing this at all.

The current idea forward is to collect all deploys , run them sequentially while tracking which succeeded so far.
In case that one fails, the list of succeeded ones is revoked one by one calling a to-be-added `revoke` subcommand of the activation binary.
 
Do you think there are more pain points we should be aware about?

> One additional implementation concern I have...

The current implementation would allow that indeed
the alternative i see would be to offer a `--profiles` flag that that's used to specify the profiles of a given flake. although then i guess the flake given should not contain a profile by itself.

ysndr:
When it comes to plain deployment of profiles, our (limited) trials show that the current implementation is capable of doing that!

ysndr:
@notgne2
We are conducting some final tests and clean up the code a bit but the tasks specified in the original comment should be implemented now, do you mind have a look on it?

While testing some questions came up:
1. when using password based authorization the magic rollbacks did not work well as ssh sessions seem to get a bit out of sync this way. The lock file delete operation runs before the lock file is created so the actual loc file is never deleted afterwards so the activation fails.
2. when doing rollbacks on previously successful nodes its a bit unclear when and if to rollback individual nodes especially with `auto-rollback` configured to `false` in the profile description or command line.
   
   our current solution to this is to  rollback all recent profiles unless `--auto-rollbacks false`  or `--auto-rollbacks` is unspecified and the profile description disabled auto-rollbacks.

"1." also happened using the unmodified version of `deploy-rs` so might be worth opening a separate issue for

ysndr:
Is there anything more required by us or is this PR ready to be merged?

notgne2:
I will look into improving password authentication, there seems to be quite a few issues with it.

My thoughts are no error-triggered rollbacks should take place with `--auto-rollback false`, and no "magic" rollbacks should take place with `--magic-rollback false`. `--revoke-succeeded` should toggle the behaviour of rolling back other profiles if a failure is encountered, with it set to `false`, the only rollbacks that should take place are per-profile as per the old functionality (and still respecting the afforementioned flags).

That sounded a bit confusing to me, so I'll give some examples in case it's confusing to you too, written assuming `.#computer2.system` is broken and will fail, but no others will.

- `deploy --auto-rollback false --magic-rollback false .#computer1.system .#computer2.system .#computer2.myservice`
  No rollbacks should take place at all, the program will quit after the failure and thus never deploy `.#computer2.myservice`

- `deploy --revoke-succeeded false .#computer1.system .#computer2.system .#computer2.myservice`
  `.#computer2.system` should be rolled back (but not `.#computer1.system`), and the program should quit after the failure and thus never deploy `.#computer2.myservice`

- `deploy --revoke-succeeded true .#computer1.system .#computer2.system .#computer2.myservice`
  `#computer2.system` should be rolled back _and_ `#computer1.system` should be rolled back, then the program should quit before `.#computer2.myservice`
  `--revoke-succeeded true` should also be the default, so specifying it shouldn't be neccesary

These should all still function without specifying the profiles too, so it should be the same when substituting `.#computer1.system .#computer2.system .#computer2.myservice` for just `.` in any of these examples, assuming that those are the only profiles provided in the flake and they are in the same order.

I haven't fully looked at the code or tested yet, but it seems like this is already how you have implemented it, so in that case there's nothing to worry about, I'm mostly summarizing it for reference.

notgne2:
One other thing I noticed right now, is the addition of `--flakes` as opposed to specifying multiple flake paths without that argument, I'm not sure what the reason is for that, if the goal were to omit the flake repo component (i.e. `deploy . --flakes computer1.system computer2.system`, though `--profiles` would probably make more sense as an argument name for that) then it would make more sense, but it doesn't look like that's how it's implemented, and I don't think it needs to either.

I mentioned earlier I wasn't sure if I want to allow evaluating multiple flake files, but now that I have thought about it more, I think I'm actually ok with that. The only issue I can think of is if we add `revoke_succeeded` to GenericSettings so that you can specify it in the flake, that would get a little difficult to resolve (as in, what flake would we trust if they disagree?), luckily I don't think adding this option is too important, so it isn't a big concern.

README.md
```diff
@@ -14,18 +14,26 @@ A Simple, multi-profile Nix-flake deploy tool.
...
+
+Any "extra" arguments will be passed into the Nix calls, so for instance to deploy an impure profile, you may use `deploy . -- --impure` (note the explicit flake path is necessary for doing this).
 
 You can try out this tool easily with `nix run`:
 - `nix run github:serokell/deploy-rs your-flake`
 
-Any "extra" arguments will be passed into the Nix calls, so for instance to deploy an impure profile, you may use `deploy . -- --impure` (note the explicit flake path is necessary for doing this).
+In you want to deploy multiple flakes or a subset of profiles with one invocation, instead of calling `deploy <flake>` you can issue `deploy --flake <flake> [<flake> ...]` where `<flake>` is supposed to take the same format as discussed before.
+
+Running in this mode, if any of the deploys fails, the deploy fill be aborted and all successful deploys revoked. `--revoke-succeeded false` can be used to override this behavior, otherwise the `auto-rollback` argument takes precedent.
```
notgne2, line 30: It might be good to explain what "revoke" means in this context here, also typo
```suggestion
Running in this mode, if any of the deploys fails, the deploy will be aborted and all successful deploys revoked. `--revoke-succeeded false` can be used to override this behavior, otherwise the `auto-rollback` argument takes precedent.
```

ysndr:
> My thoughts are no error-triggered rollbacks should take place with --auto-rollback false, and no "magic" rollbacks should take place with --magic-rollback false. --revoke-succeeded should toggle the behaviour of rolling back other profiles if a failure is encountered, with it set to false, the only rollbacks that should take place are per-profile as per the old functionality (and still respecting the afforementioned flags).

That is the functionality we built in as far as I am aware

ysndr:
> One other thing I noticed right now, is the addition of --flakes as opposed to specifying multiple flake paths without that argument, I'm not sure what the reason is for that, if the goal were to omit the flake repo component (i.e. deploy . --flakes computer1.system computer2.system, though --profiles would probably make more sense as an argument name for that) then it would make more sense, but it doesn't look like that's how it's implemented, and I don't think it needs to either.

The reason for the new flag is that there can only be one "list of arguments" without a flag. That one is already in use for the extra arguments to be passed to nix. `deploy --flake <f1> [<f*>]` and `deploy <f>`  are mutually exclusive.

README.md
ysndr: > It might be good to explain what "revoke" means in this context here, also typo

Thinking about it, we should probably call it `rollback` to be in line with the existing terminology

ysndr:
Rebased onto new master.

ysndr:
> it also seems to be missing sudo_cmd logic, which I assume will make it unable to revoke deployments as any user other than the SSH user.

I added a sudo invocation as with the other commands

ysndr:
> There seems to be some issues with the revoke subcommand, the builder for it is ommitting --temp-path which makes sense, though the command schema hasn't been adapted to support it so results in an error when revoking a deployment

The temp_path is just used for the lock file right?
I don't see which "command schema hasn't been adapted" that could result in an error.

notgne2:
What I mean is that in `src/bin/activate.rs` we have
```rust
    /// Path for any temporary files that may be needed during activation
    #[clap(long)]
    temp_path: String,
```
(note: not `Option<String>`) in `Opts` for the top level command, so regardless of the subcommand, it will fail during runtime if it is not supplied, and `build_revoke_command` doesn't supply it

ysndr:
Oh I see, that's an issue indeed.
I'll fix that later today

ysndr:
I opted to give a default value for that field, this way it wont error late during runtime if no value was given in case of the `activate` and `wait` command.
Alternatively, either said later error or providing an empty path with the revoke command are options I can think of right now

notgne2:
I would rather not give that a default value within the activation binary, and it doesn't look like `revoke` needs it, so I think it makes more sense to take it out of `Opts`, and put it in `ActivateOpts` and `WaitOpts` instead (might need to update the other command builders to have it in the right order after this though)

ysndr:
Moved temp_path to the respective opts structs and adapted the builder functions

balsoft:
Hey @ysndr, sorry for the long wait. @notgne2 Do you want me to give this a review and merge if everything is fine, or will you do this yourself?

notgne2:
@balsoft yes that would be fine, I just haven't had the time to test it properly

ysndr:
I wonder if there is a reliable way to automatically test deploy-rs
I was starting to write up something earlier but had to drop it due to time constrains

notgne2:
I had also spent a day or two trying to make a nice testing tool for deploy-rs, but I didn't manage to make a nice solution and didn't continue working on any of the uglier ones, though I'm making it the next major task to finish something as it would make further development so much easier. Right now I'm still doing tests by hand by grabbing an old laptop/homeserver I have spare, deploying a bunch of profiles to it, and hoping I didn't forget something to test each time, which is annoying and time consuming

src/bin/deploy.rs
```diff
@@ -538,7 +612,18 @@ async fn run() -> Result<(), RunError> {
         deploy::LoggerType::Deploy,
     )?;
 
-    let deploy_flake = deploy::parse_flake(opts.flake.as_str())?;
+    let deploys = opts.clone().targets.unwrap_or_else(|| {
+        opts.clone()
+            .target
+            .or_else(|| Some(".".to_string()))
+            .map(|f| vec![f])
+            .unwrap()
```
balsoft, line 620:
```suggestion
            .unwrap_or(Some(vec![".".to_string()]))
```

src/lib.rs
```diff
@@ -59,6 +59,22 @@ pub fn logger_formatter_wait(
...
+pub fn logger_formatter_revoke(
+    w: &mut dyn std::io::Write,
+    _now: &mut DeferredNow,
+    record: &Record,
+) -> Result<(), std::io::Error> {
+    let level = record.level();
+
+    write!(
+        w,
+        "🕙 {} [revoke] [{}] {}",
```
balsoft, line 71: Maybe ↩️ ?

balsoft (approved):
Looks absolutely great! Consider these suggestions, and also consider adding your authorship information to SPDX headers. You are free to ignore those, I'll merge it anyways, since the suggestions are minor.

balsoft:
Great, finally please clean up the commit history a bit (just squash everything if you're not ready to keep it clean) and rebase on master.

ysndr:
Rebasing it more granularly caused some trouble, so there we go one big commit.

balsoft:
@ysndr, thank you very much for this, and apologies for the long wait. Happy nixing!

ysndr:
@balsoft @notgne2 thanks for merging this it was a fun experience :)

PR #81, opened 2021-03-31
Add dry-activate

wigust:
```
🚀 ℹ️ [deploy] [INFO] Activating profile `system` for node `web34`
would stop the following units: docker-ssh-sup-room.service
would start the following units: docker-ssh-sup-room.service
🚀 ℹ️ [deploy] [INFO] Success activating, done!
```


notgne2:
This is a nice feature to have, and I like the way it's implemented too. One thing I don't understand though is the purpose of the `dryActivate` config option, it makes a lot more sense to me as only a CLI argument

wigust:
```
$ deploy --dry-activate -s .#router2 -- --print-build-logs
🚀 ℹ️ [deploy] [INFO] Evaluating flake in .
warning: Git tree '/home/oleg/nixos/router' is dirty
🚀 ℹ️ [deploy] [INFO] The following profiles are going to be deployed:
[router2.system]
user = "root"
ssh_user = "root"
path = "/nix/store/41wlil15dpgzzshrskh3qm50xqrinqnx-activatable-nixos-system-router2-21.05.20210303.a2b0ea6"
hostname = "router2"
ssh_opts = []

🚀 ℹ️ [deploy] [INFO] Building profile `system` for node `router2`
warning: Git tree '/home/oleg/nixos/router' is dirty
activatable-nixos-system-router2> created 21 symlinks in user environment
warning: unknown setting 'experimental-features'
warning: unknown setting 'experimental-features'

$ deploy -s .#router2 -- --print-build-logs
🚀 ℹ️ [deploy] [INFO] Evaluating flake in .
warning: Git tree '/home/oleg/nixos/router' is dirty
🚀 ℹ️ [deploy] [INFO] The following profiles are going to be deployed:
[router2.system]
user = "root"
ssh_user = "root"
path = "/nix/store/41wlil15dpgzzshrskh3qm50xqrinqnx-activatable-nixos-system-router2-21.05.20210303.a2b0ea6"
hostname = "router2"
ssh_opts = []

🚀 ℹ️ [deploy] [INFO] Building profile `system` for node `router2`
warning: Git tree '/home/oleg/nixos/router' is dirty
warning: unknown setting 'experimental-features'
🚀 ℹ️ [deploy] [INFO] Activating profile `system` for node `router2`
warning: unknown setting 'experimental-features'
updating GRUB 2 menu...
activating the configuration...
setting up /etc...
reloading user units for root...
reloading user units for eng...
setting up tmpfiles
⭐ ℹ️ [activate] [INFO] Activation succeeded!
🚀 ℹ️ [deploy] [INFO] Success activating, done!
```

wigust:
I applied your suggestions.

notgne2:
I might be reading this wrong, but it looks like after a `--dry-activate` invocation, the profile path (i.e. `/nix/var/nix/profiles/system`) will be left pointing to the closure used for running `dry-activate`, which could have dangerous unintended side-effects. There's a few ways this could be dealt with:

1. Don't apply to a path at all, instead invoke `deploy-rs-activate` with `$PROFILE` (and `$PWD`) pointing to the Nix store closure, this is simple and I like this the most, but I'm not sure if it would work or if NixOS would be upset
2. Apply to a different path, such as maybe appending `_dry` or something to the end of `profile_path`, and then deleting the profile afterwards, this seems a bit ugly especially given it could technically conflict with another path, but is otherwise a lot like 1., but working around any potential issues it may cause
3. Rolling back afterwards, this is the simplest and seemingly the sanest (as long as implemented well, such that no errors or lack thereof can prevent this rollback from occuring), but at the same time could still result in issues if anything is actively referencing the profile path while the dry-activate deployment is running.

I'm also not entirely sure what I would consider best, so I'm tagging in @balsoft and @mkaito for opinions too

balsoft:
I think option (1) is the best one. NixOS seems to be fine dry-activating from any location, and it doesn't introduce any ugliness. A big potential issue is that various activation scripts need to handle dry-activation correctly. One way to do this would be to change `activate.custom` to accept another optional argument (via `__functor` and `__toString`) that would be the dry-activation script, and in case it's not provided just print the contents of activation script; alternatively, just warn users that `dry-activate` may result in actual activation and data loss unless it is implemented correctly by the activation script.

notgne2:
I'm not sure how the optional argument system you propose could be implemented, but if at all possible that sounds preferable to me, as it sounds very easy to accidentally run `deploy --dry-activate` (as opposed to `deploy .#mycomputer.system --dry-activate`) and end up activating things by accident

balsoft:
It's simple (but not easy, although now I have already spent the time to think about so it's easy for me I guess) to implement, see https://github.com/serokell/vault-secrets/blob/master/scripts/vault-push-approles.nix for an example

wigust:
I applied method 1 and @balsoft suggestions.  From my research `/nix/var/nix/profiles/system` (pointing to a new `/nix/var/nix/profiles/system-X-link` system profile) is modified by `nix-env` invokated in `bin/activate.rs`, which we don't need to invoke if `--dry-activate` is specified.

flake.nix
```diff
@@ -59,34 +59,52 @@
...
+                          name = base.name + "-activate-path";
+                          text = ''
+                            #!${pkgs.runtimeShell}
+                            set -euo pipefail
+
+                            if [[ $DRY_ACTIVATE == "1" ]]
+                            then
+                                if ${pkgs.lib.boolToString hasDryActivate}
+                                then
+                                    ${if hasDryActivate then customSelf.dryActivate else ":"}
```
notgne2, line 83: Like @balsoft said, it might be nicer to have it print the contents of the activation script as a default instead of just noop.

src/bin/activate.rs
```diff
@@ -348,70 +352,76 @@ pub async fn activate(
...
+    }
 
     debug!("Running activation script");
 
     let activate_status = match Command::new(format!("{}/deploy-rs-activate", profile_path))
-        .env("PROFILE", &profile_path)
-        .current_dir(&profile_path)
+        .env("PROFILE", &closure)
+        .env("DRY_ACTIVATE", if dry_activate { "1" } else { "0" })
+        .current_dir(&closure)
```
notgne2, line 383: I don't think `closure` should be used in place of `profile_path` unconditionally, only when `dry_activate` is true. Alternatively, we could also just add a different environment variable, i.e. `$CLOSURE`, and have that be used within the dry-activate scripts instead of `$PROFILE`

src/deploy.rs
```diff
@@ -227,6 +237,8 @@ pub async fn deploy_profile(
 
     let auto_rollback = deploy_data.merged_settings.auto_rollback.unwrap_or(true);
 
+    let dry_activate = dry_activate;
```
notgne2, line 240: ?

src/deploy.rs
```diff
@@ -267,7 +280,9 @@ pub async fn deploy_profile(
             a => return Err(DeployProfileError::SSHActivateExitError(a)),
         };
 
-        info!("Success activating, done!");
+        if !dry_activate {
+            info!("Success activating, done!");
```
notgne2, line 284: Might be worth adding a dry-activate equivilent of this (and the "Activating profile" message above) instead of just removing

flake.nix
```diff
@@ -59,34 +59,52 @@
...
+                            exec ${self.defaultPackage."${system}"}/bin/activate "$@"
+                          '';
+                          executable = true;
+                          destination = "/activate-rs";
+                        })
+                      ];
+                  };
+              };
+
+            nixos = base: (custom // { dryActivate = "$PROFILE/bin/switch-to-configuration dry-activate"; }) base.config.system.build.toplevel ''
```
balsoft, line 107: This is not quite the interface I had in mind, but ok :) I was thinking more along the lines of just being able to `custom profile activation-script dry-activation script`

This should be doable with a `__functor//__toString` set but if you don't want to, this interface is good enough!

notgne2: I think attrsset instead of positional arguments is better in case we ever want to augment it futher in the future

wigust: @balsoft I don't understand your propose well enough to apply it, sorry.

@notgne2 I applied all your suggestions.

src/bin/activate.rs
```diff
@@ -348,70 +352,76 @@ pub async fn activate(
...
             }
-            return Err(ActivateError::SetProfileExitError(a));
-        }
-    };
+        };
+    }
 
     debug!("Running activation script");
 
     let activate_status = match Command::new(format!("{}/deploy-rs-activate", profile_path))
```
notgne2, line 380:
```suggestion
    let activate_status = match Command::new(format!("{}/deploy-rs-activate", closure))
```

notgne2:
Urgh, GitHub's code suggestion thing sucks and I don't have push permissions, here's a patch for an error I encountered

```
From 49226738ae14c28370060e115c12693ccb8cf757 Mon Sep 17 00:00:00 2001
From: notgne2 <gen2@gen2.space>
Date: Thu, 29 Apr 2021 17:34:24 -0700
Subject: [PATCH] Use correct path to `deploy-rs-activate` binary when using
 `dry_activate`

---
 src/bin/activate.rs | 8 +++++---
 1 file changed, 5 insertions(+), 3 deletions(-)

diff --git a/src/bin/activate.rs b/src/bin/activate.rs
index da53174..d17f3a8 100644
--- a/src/bin/activate.rs
+++ b/src/bin/activate.rs
@@ -377,10 +377,12 @@ pub async fn activate(
 
     debug!("Running activation script");
 
-    let activate_status = match Command::new(format!("{}/deploy-rs-activate", profile_path))
-        .env("PROFILE", if dry_activate { &closure } else { &profile_path })
+    let activation_location = if dry_activate { &closure } else { &profile_path };
+
+    let activate_status = match Command::new(format!("{}/deploy-rs-activate", activation_location))
+        .env("PROFILE", activation_location)
         .env("DRY_ACTIVATE", if dry_activate { "1" } else { "0" })
-        .current_dir(if dry_activate { &closure } else { &profile_path })
+        .current_dir(activation_location)
         .status()
         .await
         .map_err(ActivateError::RunActivateError)
-- 
2.31.1
```

wigust:
@notgne2 your patch 49226738ae14c28370060e115c12693ccb8cf757 applied and pushed.

jared-w:
Here's a fun one that I also ran into.

If your profile is new enough that what's on the machine doesn't have this feature and what's on the host does, it's possible to get into a situation where a failure in the profile activation will cause the rollback to bork because of a missing `DRY_ACTIVATE` variable.

flake.nix
```diff
@@ -59,34 +59,48 @@
...
+                    paths =
+                      [
+                        base
+                        (pkgs.writeTextFile {
+                          name = base.name + "-activate-path";
+                          text = ''
+                            #!${pkgs.runtimeShell}
+                            set -euo pipefail
+
+                            if [[ $DRY_ACTIVATE == "1" ]]
```
balsoft, line 76: Oh yeah, right.

@notgne2

```suggestion
                            if [[ ${DRY_ACTIVATE:-} == "1" ]]
```

jared-w: I'm not sure `${DRY_ACTIVATE:-}` will work in the presence of `set -u`? If it does, then great. But yeah, I ran into that briefly and was losing my mind for a minute before I realized the full path of the rollback binary was local to the profile

balsoft: Yeah, right, also needs to be quoted.

PR #82, opened 2021-04-05
Add helper activation function for home-manager

### LINKED ISSUE #43: Activation preset for home-manager
Despite intentionally remaining profile-agnostic, deploy-rs provides support for NixOS in the form of `activate.nixos`, I would like to provide a similar option for home-manager deployments, this is something I know to be possible as I use it in my own deployments, though last time I checked required modifying some of home-manager's code.

notgne2:
This resolves #43, though I would also like to add example usage, and we may want to change out `hm` for `home-manager`, I'm not sure what one is best

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -112,6 +112,8 @@
               "sed -i '/^default /d' ${efi.efiSysMountPoint}/loader/loader.conf"}
             '';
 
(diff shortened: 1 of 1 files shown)


balsoft (approved):

notgne2:
(added a comment linking to home-manager, as with the name `hm` it's not immediately clear)

PR #90, opened 2021-05-18
Fix auto-rollback with dry_activate

balsoft:
- Fix rollback to old profiles
- flake.nix: clean up dry-activation logic

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -73,13 +73,9 @@
                             #!${pkgs.runtimeShell}
                             set -euo pipefail
 
-                            if [[ "x''${DRY_ACTIVATE:-}" == "x1" ]]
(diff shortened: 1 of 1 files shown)


flake.nix
```diff
@@ -73,7 +73,7 @@
                             #!${pkgs.runtimeShell}
                             set -euo pipefail
 
-                            if [[ $DRY_ACTIVATE == "1" ]]
+                            if [[ "x''${DRY_ACTIVATE:-}" == "x1" ]]
```
zhenyavinogradov, line 76: Why is `x` prepended here, are there cases where this won't work otherwise?

zhenyavinogradov (approved):

flake.nix
balsoft: You're right, doesn't make a difference.

PR #93, opened 2021-06-06
Mention the matrix channel in the README

balsoft:
We now have a matrix channel. To increase community engagement, we should mention it in the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,8 @@ SPDX-License-Identifier: MPL-2.0
 
 A Simple, multi-profile Nix-flake deploy tool.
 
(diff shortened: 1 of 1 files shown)


kirelagin:
We are waiting for it to be aliased in nixos.org though, aren't we?

kirelagin:
OTOH, updating the address in the README again later is free, so why not.

kirelagin (approved):

balsoft:
> We are waiting for it to be aliased in nixos.org though, aren't we?

AFAIK it's unlikely we'll actually get an alias, and besides the link will work even after we do.

PR #95, opened 2021-06-21
flake.nix: refactor as overlay

balsoft:
From #deploy-rs:matrix.org:

> <@pachumicchu:myrdd.info> We're working on the issue of creating too many nixpkgs instances in digga, and one slight hiccup we noticed is that deploy-rs has a "system-spaced" lib output(lib.x86_64.function) and therefore instantiates nixpkgs on its own. I'm thinking that it could be good to have copies of those lib functions that take pkgs as an input, so you can pass it your own instantiated nixpkgs. So lib.function = { pkgs, ... }: pkgs.doSomething.
> <@balsoft:balsoft.ru> It should be rather simple to implement, so please go ahead! Although I think it would be better to put it into overlay output.

I have decided to go ahead myself and do this. This adds an overlay, and uses that overlay to provide actual flake outputs. Then the consumers can use the existing overlays mechanism to prevent unneeded nixpkgs evaluations.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -20,37 +20,24 @@
   };
 
   outputs = { self, nixpkgs, utils, naersk, ... }:
-    utils.lib.eachDefaultSystem (system:
-      let
-        pkgs = import nixpkgs { inherit system; };
-        naersk-lib = pkgs.callPackage naersk { };
-        isDarwin = pkgs.lib.strings.hasSuffix "-darwin" system;
-        darwinOptions = pkgs.lib.optionalAttrs isDarwin {
-          nativeBuildInputs = [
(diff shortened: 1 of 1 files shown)


mkaito:
That is an obnoxious amount of horizontal scrolling.

mkaito (approved):
Actually looks good, but maybe add a usage example somewhere?

blaggacao (approved):

blaggacao:
Once `divnix/nixpkgs.lib` is under `nix-community`, I guess I'll open another PR for making use of it for all things `lib`.

Pacman99:
pachumicchu from matrix here, thank you for implementing this so quickly! This looks great :+1:

PR #96, opened 2021-06-25
Evaluate only once, and then only realise

balsoft:
> Add rustfmt and rust-analyzer to devShell

I'm tired of adding this manually before I hack on this

> Evaluate only once, and then only realise

A dumb hack that uses `nix show-derivation` to get the deriver of a not-yet-existing path, and then builds that derivation using `nix build`/`nix-build`.

### DIFF
--- a/src/push.rs
+++ b/src/push.rs
@@ -3,6 +3,7 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use log::{debug, info};
+use std::collections::HashMap;
 use std::path::Path;
 use std::process::Stdio;
 use thiserror::Error;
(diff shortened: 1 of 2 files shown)


src/push.rs
```diff
@@ -49,22 +58,37 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
...
+    let mut show_derivation = Command::new("nix");
+
+    show_derivation
+        .arg("show-derivation")
+        .arg(&data.deploy_data.profile.profile_settings.path);
+
+    let derivation_info = show_derivation
+        .output()
+        .await
+        .map_err(PushProfileError::ShowDerivationError)?
```
notgne2, line 71: I think we should check the exit code of this too

src/push.rs
```diff
@@ -3,13 +3,22 @@
...
 use std::process::Stdio;
 use thiserror::Error;
 use tokio::process::Command;
 
 #[derive(Error, Debug)]
 pub enum PushProfileError {
+    #[error("Failed to run Nix show-derivation command: {0}")]
+    ShowDerivationError(std::io::Error),
+    #[error("Nix show-derivation command output contained an invalid UTF-8 sequence: {0}")]
+    ShowDerivationUtf8Error(#[from] std::str::Utf8Error),
```
notgne2, line 17: I think I've grown to prefer `map_err` over `#[from]`, but don't mind

src/push.rs
```diff
@@ -49,22 +58,37 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
         data.deploy_data.profile_name, data.deploy_data.node_name
     );
 
+    // `nix-store --query --deriver` doesn't work on invalid paths, so we parse output of show-derivation :(
+    let mut show_derivation = Command::new("nix");
```
notgne2, line 62: maybe `show_derivation_command` would be nicer

src/push.rs
```diff
@@ -49,22 +58,37 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
         data.deploy_data.profile_name, data.deploy_data.node_name
     );
```
notgne2, line 60: Maybe we should update/re-arrange the log messages to reflect the new behavior

notgne2 (commented):
Here is my silly nitpicks, I'll start testing it too now

notgne2 (approved):

PR #102, opened 2021-07-11
Skip checks by default

balsoft:
Running checks by default is annoying and makes iterative deployment a nuiscance. Disable checks by default, deprecate `--skip-checks` and add `--check` to enable checking.


mkaito:
Seems potentially controversial. Add a setting in the flake spec?

notgne2 (approved):
eh sure why not

teto:
software should be designed for usage, not for development. I think the current default is better (just broken because it evaluates everything instead of just the current target).

PR #104, opened 2021-07-15
Use as library

nrdxp:
Turns out, it wasn't too difficult to turn deploy-rs into a library for use as a subcommand in other rust binaries. Not sure if this is a use case you wish to support, which is why it's in draft for now, but it works [amazingly well](https://github.com/input-output-hk/bitte-cli/pull/18).


balsoft (approved):
Looks good, I don't have any objections.

balsoft:
Actually,
```
   Doc-tests deploy
error[E0659]: `deploy` is ambiguous (name vs any other name during import resolution)
  --> /build/source/src/cli.rs:13:5
   |
13 | use deploy::{DeployFlake, ParseFlakeError};
   |     ^^^^^^ ambiguous name
   |
   = note: `deploy` could refer to a crate passed with `--extern`
   = help: use `::deploy` to refer to this crate unambiguously
note: `deploy` could also refer to the module imported here
  --> /build/source/src/cli.rs:11:5
   |
11 | use crate as deploy;
   |     ^^^^^^^^^^^^^^^
   = help: use `self::deploy` to refer to this module unambiguously

error: aborting due to previous error

For more information about this error, try `rustc --explain E0659`.
error: test failed, to rerun pass '--doc'
```

nrdxp:
Should be fixed now. Good catch @balsoft
And it's a good reminder to always run the tests before opening a PR :sweat_smile:

balsoft (approved):

nrdxp:
Now that this is merged, do you think it would be worth publishing on crates.io @balsoft?

balsoft:
We'll think about it, thanks for the suggestion!

PR #109, opened 2021-08-04
General code cleanup

notgne2:
This doesn't really contain any structural changes, but cleans up many smaller (mostly clippy motivated) issues. After this I will likely make a PR for some of the more project-specific small changes, followed by one with some of the larger diff changes.

### DIFF
--- a/src/bin/activate.rs
+++ b/src/bin/activate.rs
@@ -95,23 +95,23 @@ struct RevokeOpts {
 #[derive(Error, Debug)]
 pub enum DeactivateError {
     #[error("Failed to execute the rollback command: {0}")]
-    RollbackError(std::io::Error),
+    Rollback(std::io::Error),
     #[error("The rollback resulted in a bad exit code: {0:?}")]
-    RollbackExitError(Option<i32>),
+    RollbackExit(Option<i32>),
     #[error("Failed to run command for listing generations: {0}")]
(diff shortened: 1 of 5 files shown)


blaggacao:
I suggest this being a boolean, instead of Option:

https://github.com/serokell/deploy-rs/blob/f53d8b92f79f5662ecd8357e72a5bb6e880f1cb4/src/bin/deploy.rs#L68

it is awkward to set `--fast-connection true` (instead of `--fast-connection`)

blaggacao:
I also found the the data structures w.r.t. `DeployData` & `DeployRefs` probably could suffer a cleanup, too?

It looks like they at the root of semantic inconsistencies like:

https://github.com/serokell/deploy-rs/blob/d72174307d5b88ec24cc2e69e875228fe3d642ed/src/deploy.rs#L294

and

https://github.com/serokell/deploy-rs/blob/d72174307d5b88ec24cc2e69e875228fe3d642ed/src/deploy.rs#L427

balsoft (approved):

notgne2:
Merged in e5546f9c2503c26d175f08a81fc0a0f330be4cbe

blaggacao:
Nice, I'll rebase today all my stuff or should we _also_ do the renaming, first? Just let me know, since it'll probably take an hour or so.

PR #110, opened 2021-08-04
Rename deploy-rs to yeet

notgne2:
This isn't all the things we will need to update if we move forward with this change, to list a few more

- logo
- more aliasing
- cross compability with pre-change (like `deploy-rs-activate` stuff)
- more documention or API naming where `deploy` refers to the tool rather than the action

and we also might want to think if this is the time to rename `activate` to something too.

But this isn't a certain decision yet, this PR is also up for us to collect feedback on the potential name change. For more context we have discussions about it on our Matrix channel [#deploy-rs:nixos.org](https://matrix.to/#/#deploy-rs:nixos.org).


balsoft:
I don't particularly care either way, but `yeet` is definitely shorter, easier to type, and more memorable. However, it gives the project a memy vibe, which may not be desirable in some circumstances.

06kellyjac:
yeet is certainly memeworthy. Maybe `trebuchet` with the short hand being `treb`? still leans into the *yeet* throwing angle

cab404 (changes requested):
Maybe add some references to previous name, so it would be easier to find by it?

*also known by it's former name "deploy-rs"*

alxpettit:
_For all sad words of tongue and pen, the saddest are these, “It might have been." -- John Greenleaf_

PR #112, opened 2021-08-05
Add default flake and node from environment values

blaggacao:
Users can set:
DEPLOY_RS_DEFAULT_FLAKE_ROOT
DEPLOY_RS_DEFAULT_NODE

If set, targets that are not paths (starting with '.'|'/') are suffixed
to those defaults.

This affords users a shorthand to their preferred flake [and node]
derived from context.


blaggacao:
I'm closing since after discussion with @balsoft , it might be out of scope. At least at the moment.

PR #113, opened 2021-08-06
add cmd completion

blaggacao:
- Add default flake and node from environment values
- Add initial completions support


src/cli.rs
```diff
@@ -18,16 +20,31 @@ use std::process::Stdio;
...
-#[clap(version = "1.0", author = "Serokell <https://serokell.io/>")]
+#[clap(name = "deploy", version = "1.0", author = "Serokell <https://serokell.io/>")]
 pub struct Opts {
+    /// If provided, outputs the completion file for given shell
+    #[clap(long = "generate", arg_enum)]
+    generator: Option<GeneratorChoice>,
+
     /// The flake to deploy
-    #[clap(group = "deploy")]
+    #[clap(group = "deploy", value_hint = ValueHint::DirPath)]
```
blaggacao, line 43: Here, prospectively a dynamic value hint would be useful.

notgne2:
I would like to have this, but this implementation looks excessively verbose. If this is something difficult to do with clap I don't mind switching to another argument parser if there is one that can handle this better (in fact, there is other issues I have with it too), and if not maybe we can try to help clap get native completion support, or at least offload more of the boilerplate-esque code.

I've also never even heard of clap_generate before, so explaining more of what is being done in this PR would be appreciated at least for documentation sake.

blaggacao:
The main issue that I currently see with clap is its lack for dynamic completion support. On the other hand, it is my (first) impression, that clap seems to be the most advanced and feature complete lib in the rust ecosystem.

[`clap_generate`](https://github.com/clap-rs/clap/tree/master/clap_generate) effectively (only) holds the code for the generation of each shell's command completion scripts.

I still have to think about what we could usefully do with this PR. While simple command completion is nice, the real power would come from dynamic command completion that is capable of suggesting nodes and profiles for any given target flake.

PR #116, opened 2021-08-10
Expand environmental variables in sshOpts

nrdxp:
For now only bracket style expansion is enabled, i.e. `${MY_VAR}`. I had issues with expanding `$MY_VAR` in a path string, and so excluded that usecase to avoid errors. One can specify brackets in a Nix string by escaping the `$`, e.g. `"\${NOT_EXPANDED_BY_NIX}"`.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -136,6 +136,7 @@ name = "deploy-rs"
 version = "0.1.0"
 dependencies = [
  "clap",
+ "envmnt",
  "flexi_logger",
  "fork",
  "futures-util",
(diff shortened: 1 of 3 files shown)


notgne2 (approved):
LGTM, but I would like to know more about the issues with the prefix style

balsoft:
> This doesn't work in '' stype strings and requires a bit more magic:
> ''${"\${NOT_EXPANDED_BY_NIX}"}''

Actually, just
```
nix-repl> ''''${NOT_EXPANDED_BY_NIX}''
"${NOT_EXPANDED_BY_NIX}"
```

nrdxp:
@notgne2, it may just be the library I used, but I have a shell that detects the flake root directory using `git rev-parse --show-toplevel` when trying to set just `"$FLAKE_ROOT/path/to/key"` it expanded to an empty string somehow :shrug: 

There is probably a similar crate, or maybe I could look at the source, and just reimplement a more minimal version or something.

@balsoft, thanks for the tip!

blaggacao:
Maybe this function or part of it can also be hammered into shape to be useful for parsing the cli string into a Vector via the [`parse(try_from_str = parse_key_val)`](https://github.com/clap-rs/clap/blob/69d75949dfe0c0e7e497ca7f167e86a98758ebd8/clap_derive/examples/keyvalue.rs#L29) attribute.

That's the last piece missing for homogenization of data structures.

PR #128, opened 2021-09-15
flake: replace naersk with buildRustPackage

lovesegfault:
We weren't using any features of naersk that weren't covered by
buildRustPackage, so let's just use the tooling already in nixpkgs for Rust
builds and get rid of a flake input.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -16,27 +16,6 @@
         "type": "github"
       }
     },
-    "naersk": {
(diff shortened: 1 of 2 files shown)


balsoft:
Looks pretty good! Did you check that the resulting executable works well?

balsoft (approved):
Ok, I confirmed that it works for our use-cases.

PR #135, opened 2021-11-02
chore: cleanup root and examples

06kellyjac:
Currently the root and especially examples are littered with `.license` files

I've removed them and replaced them with dep5 configuration since you already have the dep5 file.

```
$ reuse lint
# SUMMARY

* Bad licenses:
* Deprecated licenses:
* Licenses without file extension:
* Missing licenses:
* Unused licenses:
* Used licenses: MPL-2.0
* Read errors: 0
* Files with copyright information: 30 / 30
* Files with license information: 30 / 30

Congratulations! Your project is compliant with version 3.0 of the REUSE Specification :-)
```

This has the benefit of cleaning up the root and also making the example dirs look a lot nicer

You could probably remove the headers from the files in the examples if you'd like

**Note:** I also symlinked the LICENSE in the root to the LICENSES dir. This change is probably controversial so I can drop that if you'd like


notgne2:
Huh that's pretty neat, I guess I'll have to read more about different ways to utilize `reuse` then, I was under the impression headers were the only way.

Though as for the symlink, I don't like it but not at a conceptual level, instead because I feel like it should be the other way round (`LICENSE` symlinks to `LICENSES/MPL-2.0.txt`) but I think GitHub did not recognize it as a license file (maybe they fixed this though?), also the symlink you included is not a relative symlink so instead references your personal directory.

06kellyjac:
I've fixed the link and swapped it to link LICENSES/MPL-2.0.txt to LICENSE

It seems to work alright in GH ui now and `reuse` is still happy with it but I'm pretty sure some license detectors don't like the /LICENSE being a link but don't care about /LICENSES/blah

image

image

---

For the headers I'd say it's best for normal code where it's easy to just add header comments but I do prefer this format for files like lockfiles etc

06kellyjac:
I've rebased on the changes in master

PR #141, opened 2021-12-16
Client-Server Split RFC

MagicRB:
Signed-off-by: Magic_RB <magic_rb@redalder.org>

You can find the text of this RFC [here](https://github.com/MagicRB/deploy-rs/blob/magic_rb/split-rfc/proposal.md).

### DIFF
--- a/proposal.md
+++ b/proposal.md
@@ -1,0 +1,135 @@
+# Client-Server Split
+
+## What is this?
+
+This is an RFC, for a major change in design. I propose to split `deploy-rs` into two fully self standing executables, a
+server and a client. This would allow for additional flexibility and some issues are impossible to solve nicely with the
+current design.
+
+## The design
+
+I propose to split `deploy-rs` into a server and a client, `deploy` and `activate` respectively. Basically, while now
+`activate` is just a helper, this RFC proposes to make it a self standing program. The basic jist is that the two
+components would communicate over a open ssh connection, using plain REST (this is important later). That would allow us
+bidirectional flow of data between the two components which greatly increases the things that can be implemented.
+
+## Benefits
+
+Fixes of sudo and ssh password login, more possibilities like better error reporting. A daemon mode and limiting the
+number of ssh connections. Parallelism and plugins/secrets (maybe).
+
+### Daemon
+
+The server component could be made in such a way that it running as a daemon listening over ssh would be equal
+to running in daemon mode and listening on TCP/IP. We'd just enable authentication and change over what it
+communicates. We could even call it `single-client-mode` and `multi-client-mode`.
+
+## Drawbacks
+
+```
+A simple multi-profile Nix-flake deploy tool.
+```
+could be considered false advertising and general increased complexity.
+
+## Bonus
+
+Since the `server` component will get scraped anyway why not write it in Haskell? If we stick to a REST API compatibility won't be an issue and contributors shouldn't be an issue either. There are more than enough Haskellers in the Nix community. We also all know how amazing Haskell is. Rust's compilation times are not that much better when compared to Haskell either.
+
+## Related
+### Issues
+- [x] - [Build on target server? #12](https://github.com/serokell/deploy-rs/issues/12)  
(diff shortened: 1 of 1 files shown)


balsoft:
At this point might as well rewrite it from scratch. Wouldn't be the first time in this project's life ([3] [2] [1] [0])

[3]: https://github.com/serokell/deploy-rs
[2]: https://github.com/serokell/deploy
[1]: https://github.com/serokell/stakerdao-infra/pull/4/commits/612f0a3c845a560f42212978810769d604641018
[0]: https://github.com/serokell/stakerdao-infra/blob/612f0a3c845a560f42212978810769d604641018/scripts/deploy.sh

MagicRB:
Every time we do so, it improves bit by bit, learning from past mistakes but starting with a clean slate.

balsoft:
As for whether it's actually a good idea? I don't know. When I started writing deploy, the vision I had was of a really simple and dumb script that just does one thing. It turns out even the simplest profile deployment can get complicated once you add all the rollback logic. Maybe rewriting it and splitting it up is the right step. Maybe we should take a step in the other direction and get rid of some extra features (magic-rollback, other fancy activation stuff) in the name of simplicity and reliability of the tool itself, so that there's no need for a separate `activate` binary, and then implement those features on top using a separate, optional, server. The result should be pretty much the same, but with less rewriting (since we can dumb down the existing codebase quite easily, and then implement the features in a new way on top).

notgne2:
>At this point might as well rewrite it from scratch.

in which case having a referenceable specification would be a good thing!

I also _kindof_ wanted to do a lot of rewriting already, I've been pondering some issues and limitations of magic-rollback and came to the conclusion that most of that logic has to be redesigned from scratch, and could probably convert a lot of my thoughts to a similar RFC too

proposal.md
```diff
@@ -0,0 +1,135 @@
...
+
+This is an RFC, for a major change in design. I propose to split `deploy-rs` into two fully self standing executables, a
+server and a client. This would allow for additional flexibility and some issues are impossible to solve nicely with the
+current design.
+
+## The design
+
+I propose to split `deploy-rs` into a server and a client, `deploy` and `activate` respectively. Basically, while now
+`activate` is just a helper, this RFC proposes to make it a self standing program. The basic jist is that the two
+components would communicate over a open ssh connection, using plain REST (this is important later). That would allow us
```
balsoft, line 13: We'd have to make a new ssh connection to confirm activation anyways, since that's precisely what we're confirming (the ability to ssh into the machine).

Also, we'd need to reimplement `nix copy` for this. Should not be too difficult, but something to keep in mind.

notgne2:
Rollback stuff makes it complicates yes, but at least after our current iteration of the codebase we know _how_ it makes it complicated, so if we include it again we are aware how to implement it more cleanly without creating annoying logic problems and disrupting the readability of the rest of the code

proposal.md
MagicRB: Well we'd still be down to 2 ssh connections from 3 or 4 or however many we have now

MagicRB:
> in which case having a referenceable specification would be a good thing!

If I had to describe how I'd imagine a perfect future for this RFC and deploy-rs/yeet generally is that before writing a single line of code, we go though all the code we have now, write down all the features we have and then analyze what each feature encompasses, the requirements and such. Create api specifications. Then as the last step we just write code which compiles and works on the first attempt :D.

MagicRB:
> Maybe rewriting it and splitting it up is the right step. Maybe we should take a step in the other direction and get rid of some extra features

How would that work? Because we need a smarter program on the remote machine to do cooler things, so I can't really imagine a way where we build something complex on a simpler base then we have now. Maybe plugins on both ends?

The way I imagined it is that when you run yeet in singles clients mode, it installs the server component and then after its done removes itself from the target automatically, the complex nature would be completely hidden from the user in that case.

MagicRB:
should I expand on the proposal? I'd start with a detailed functionality description and feature set.

PR #148, opened 2022-01-27
Add custom sudo command support

### LINKED ISSUE #145: support doas in replace of sudo
it would be nice to have an option to use doas as well as sudo to deploy nix configurations

afontaine:
This is useful for nodes that utilize `doas` instead of `sudo`.

---

This is my first rust anything, so I'm not sure if any of this is correct (or if I can write unit tests).

I need to add documentation still as well, but getting it up in case my rust is bad was my first priority.

Closes #145

### DIFF
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -161,6 +161,7 @@ pub struct CmdOverrides {
     pub magic_rollback: Option<bool>,
     pub temp_path: Option<String>,
     pub confirm_timeout: Option<u16>,
+    pub sudo: Option<String>,
     pub dry_activate: bool,
 }
 
@@ -350,7 +351,7 @@ impl<'a> DeployData<'a> {
         let profile_path = self.get_profile_path()?;
 
         let sudo: Option<String> = match self.merged_settings.user {
-            Some(ref user) if user != &ssh_user => Some(format!("sudo -u {}", user)),
+            Some(ref user) if user != &ssh_user => Some(format!("{} {}", self.get_sudo(), user)),
             _ => None,
         };
 
@@ -391,6 +392,13 @@ impl<'a> DeployData<'a> {
             },
         };
         Ok(profile_user)
+    }
+
+    fn get_sudo(&'a self) -> String {
+        return match self.merged_settings.sudo {
+           Some(ref x) => x.clone(),
+           None => "sudo -u".to_string()
+        };
     }
 }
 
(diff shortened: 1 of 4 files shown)


src/lib.rs
```diff
@@ -392,6 +393,14 @@ impl<'a> DeployData<'a> {
...
         Ok(profile_user)
     }
+
+    fn get_sudo(&'a self) -> Result<String, DeployDataDefsError> {
+       let sudo = match self.merged_settings.sudo {
+           Some(ref x) => x.clone(),
+           None => "sudo -u".to_string()
+        };
+        Ok(sudo)
+    }
```
afontaine, line 403: Does it make sense to provide a default value here, or are defaults hidden somewhere else?

src/lib.rs
```diff
@@ -392,6 +393,14 @@ impl<'a> DeployData<'a> {
         };
         Ok(profile_user)
     }
+
+    fn get_sudo(&'a self) -> Result<String, DeployDataDefsError> {
```
afontaine, line 397: Does this need to be a `Result` type? I don't think there can be any errors to return here :thinking:

balsoft: No, it doesn't. And if it had to be a `Result`, the `Error` type should have been different anyways :)

src/lib.rs
balsoft: This looks fine to me :)

balsoft:
This is missing some changes:
```
diff --git a/src/cli.rs b/src/cli.rs
index 61890e4..cc5a3ac 100644
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -89,6 +89,9 @@ pub struct Opts {
     /// Revoke all previously succeeded deploys when deploying multiple profiles
     #[clap(long)]
     rollback_succeeded: Option<bool>,
+    /// Which sudo command to use. Must accept at least two arguments: user name to execute commands as and the rest is the command to execute
+    #[clap(long)]
+    sudo: Option<String>,
 }
 
 /// Returns if the available Nix installation supports flakes
@@ -635,6 +638,7 @@ pub async fn run(args: Option<&ArgMatches>) -> Result<(), RunError> {
         temp_path: opts.temp_path,
         confirm_timeout: opts.confirm_timeout,
         dry_activate: opts.dry_activate,
+        sudo: opts.sudo,
     };
 
     let supports_flakes = test_flake_support().await.map_err(RunError::FlakeTest)?;
```

afontaine:
Awesome, thank you @balsoft!

I've added the CLI option (correct?), as well as un-`Result`ed the fetch command
and added a blurb in the README. Does this all look good?

balsoft (approved):

PR #158, opened 2022-05-14
Build every profile first, then push

2xsaiko:
Try to build everything first before pushing to remotes. Since the build
is more likely to fail than the upload, if there is an error the deployment
will fail sooner and before uploading any potentially unusable configuration.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -539,18 +539,27 @@ async fn run_deploy(
         print_deployment(&parts[..])?;
     }
 
-    for (deploy_flake, deploy_data, deploy_defs) in &parts {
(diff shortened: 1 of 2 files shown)


teto:
this could be done via a wrapping script, but if it gets integrated, I think it should be merged with the -remote-build flag , for instance a --build-strategy flag with values ("remote", "full", "jit" for instance)

balsoft (approved):
LGTM

Please, rebase and fix conflicts

balsoft:
`nix flake check` passes, `deploy`-ing works

balsoft:
Thanks!

PR #159, opened 2022-05-14
Update flake to support nix 2.8

xvello:
Nix 2.7 [renamed defaultApp and defaultPackage](https://discourse.nixos.org/t/nix-2-7-0-released/18072). Both the old and new names are supported in 2.7, but 2.8 has removed support for the old names, breaking the `nix run github:serokell/deploy-rs .` invocation.

Old names are kept in this PR to keep compatibility with nix 2.6, but could be removed if support of this version is not needed anymore.

PR tested with 2.6.1, 2.7.0 and 2.8.0

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -74,7 +74,7 @@
                             name = base.name + "-activate-rs";
                             text = ''
                             #!${final.runtimeShell}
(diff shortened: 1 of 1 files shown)


xvello:
Heya @balsoft, would you have a moment to review this PR please?

PR #175, opened 2022-09-14
Add option to build on the target host

### LINKED ISSUE #12: Build on target server?
Is it possible to build the flake on the targets server instead of my laptop?
I don't want to upload big build closures to my machines all the time.

PhilTaken:
This PR adds the option to build the derivation on the target server instead of locally.
The derivation will first be uploaded to the target system and then built in the remote store.
Any required external dependencies will be fetched to the target host using the target system's substituters.

Fixes #12

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -56,6 +56,10 @@ pub struct Opts {
     #[clap(short, long)]
     skip_checks: bool,
 
+    /// Build on remote host
+    #[clap(long)]
+    remote_build: bool,
+
     /// Override the SSH user with the given value
     #[clap(long)]
     ssh_user: Option<String>,
@@ -138,9 +142,7 @@ async fn check_deployment(
                 .arg(format!("let r = import {}/.; x = (if builtins.isFunction r then (r {{}}) else r); in if x ? checks then x.checks.${{builtins.currentSystem}} else {{}}", repo));
     }
 
-    for extra_arg in extra_build_args {
-        check_command.arg(extra_arg);
-    }
+    check_command.args(extra_build_args);
 
     let check_status = check_command.status().await?;
 
@@ -239,9 +241,7 @@ async fn get_deployment_data(
             .arg(format!("let r = import {}/.; in if builtins.isFunction r then (r {{}}).deploy else r.deploy", flake.repo))
     };
 
-    for extra_arg in extra_build_args {
-        c.arg(extra_arg);
-    }
+    c.args(extra_build_args);
 
     let build_child = c
         .stdout(Stdio::piped())
@@ -640,6 +640,7 @@ pub async fn run(args: Option<&ArgMatches>) -> Result<(), RunError> {
         temp_path: opts.temp_path,
         confirm_timeout: opts.confirm_timeout,
         dry_activate: opts.dry_activate,
+        remote_build: opts.remote_build,
         sudo: opts.sudo,
     };
 
--- a/src/push.rs
+++ b/src/push.rs
@@ -41,6 +41,8 @@ pub enum PushProfileError {
     Copy(std::io::Error),
     #[error("Nix copy command resulted in a bad exit code: {0:?}")]
     CopyExit(Option<i32>),
+    #[error("The remote building option is not supported when using legacy nix")]
+    RemoteBuildWithLegacyNix,
 }
 
 pub struct PushProfileData<'a> {
@@ -54,40 +56,7 @@ pub struct PushProfileData<'a> {
     pub extra_build_args: &'a [String],
 }
 
-pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileError> {
-    debug!(
-        "Finding the deriver of store path for {}",
-        &data.deploy_data.profile.profile_settings.path
-    );
-
-    // `nix-store --query --deriver` doesn't work on invalid paths, so we parse output of show-derivation :(
-    let mut show_derivation_command = Command::new("nix");
-
-    show_derivation_command
-        .arg("show-derivation")
-        .arg(&data.deploy_data.profile.profile_settings.path);
-
-    let show_derivation_output = show_derivation_command
-        .output()
-        .await
-        .map_err(PushProfileError::ShowDerivation)?;
-
-    match show_derivation_output.status.code() {
-        Some(0) => (),
-        a => return Err(PushProfileError::ShowDerivationExit(a)),
-    };
-
-    let derivation_info: HashMap<&str, serde_json::value::Value> = serde_json::from_str(
-        std::str::from_utf8(&show_derivation_output.stdout)
-            .map_err(PushProfileError::ShowDerivationUtf8)?,
-    )
-    .map_err(PushProfileError::ShowDerivationParse)?;
-
-    let derivation_name = derivation_info
-        .keys()
-        .next()
-        .ok_or(PushProfileError::ShowDerivationEmpty)?;
-
+pub async fn build_profile_locally(data: &PushProfileData<'_>, derivation_name: &str) -> Result<(), PushProfileError> {
     info!(
         "Building profile `{}` for node `{}`",
         data.deploy_data.profile_name, data.deploy_data.node_name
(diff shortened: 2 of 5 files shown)


src/cli.rs
```diff
@@ -239,9 +237,7 @@ async fn get_deployment_data(
             .arg(format!("let r = import {}/.; in if builtins.isFunction r then (r {{}}).deploy else r.deploy", flake.repo))
     };
 
-    for extra_arg in extra_build_args {
-        c.arg(extra_arg);
-    }
+    c.args(extra_build_args);
```
rvem, line 240: Can you also add `--remote-build` flag to the command line arguments of `deploy`?

src/push.rs
```diff
@@ -99,12 +99,46 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
...
     };
 
+    // copy derivation over to remote server
+    if data.supports_flakes && data.deploy_data.merged_settings.remote_build.unwrap_or(false) {
+        info!("Building on remote host");
+
+        // copy the derivation to remote host so it can be built there
+        let copy_command_status = Command::new("nix").arg("copy")
+            .arg("-s")  // fetch dependencies from substitures, not localhost
+            .arg("--to").arg(&store_address)
```
rvem, line 124: Looks like this doesn't inherit an ssh port, thus an attempt to use it may fail, e.g.
```
ssh: connect to host skat.aquarius.serokell.team port 22: Network is unreachable
error: cannot open a connection to remote store 'ssh-ng://rvem@<bla-bla>': error: unexpected end-of-file
```
While deploy configuration uses non-default ssh port

src/push.rs
```diff
@@ -99,12 +99,46 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
...
+            .arg("--derivation").arg(derivation_name)
```
rvem, line 125: Something like
```suggestion
            .arg("--derivation").arg(derivation_name)
            .env("NIX_SSHOPTS", &ssh_opts_str)
```
And moving `ssh_opts_str` definition above seems to help

rvem: Hmm, after this change deployment hangs in
```
🚀 ℹ️ [deploy] [INFO] Building on remote host
```
state. And I don't see `nix build` on a target machine :(

src/cli.rs
PhilTaken: as a "build all deployments on the respective target servers" override?
the current implementation only builds selected builds remotely, but that could be a good idea

rvem: > as a "build all deployments on the respective target servers" override?

Yes, I imagine that it might be useful when one doesn't have decent internet connection or has limited traffic

src/push.rs
```diff
@@ -99,12 +99,58 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
...
+            .env("NIX_SSHOPTS", ssh_opts_str.clone())
```
rvem, line 137: Okay, it doesn't seem to work that easily :(

```
🚀 ℹ️ [deploy] [INFO] Building profile `system` for node `<bla>`
🚀 ℹ️ [deploy] [INFO] Building on remote host
ssh: connect to host <bla-bla> port 22: Network is unreachable
error: cannot open connection to remote store 'ssh-ng://rvem@<bla-bla>': error: unexpected end-of-file
🚀 ❌ [deploy] [ERROR] Failed to push profile: Nix build command resulted in a bad exit code: Some(1)
```

src/push.rs
```diff
@@ -118,9 +164,7 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
         (false, true) => build_command.arg("--no-link"),
     };
 
-    for extra_arg in data.extra_build_args {
-        build_command.arg(extra_arg);
-    }
+    build_command.args(data.extra_build_args);
 
     let build_exit_status = build_command
         // Logging should be in stderr, this just stops the store path from printing for no reason
```
hazelweakly, line 170: When I try this out and run it on my machine, it fails on line 181

```rust
  if !Path::new(
        format!(
            "{}/deploy-rs-activate",
            data.deploy_data.profile.profile_settings.path
        )
        .as_str(),
```

because, as the system is remote and the closure was built remotely, this path is not going to exist in my store or on my system.

I'm not sure I have a _great_ solution for that, but I wanted to point that out cause I'm trying this out in anticipation of managing remote machines via macOS

rvem: Perhaps, in this case, activation should happen on the target as well :thinking:

src/push.rs
rvem: bThough, according to debug logs OpenSSH succesfully  the non-default port:
```
Authenticated to <bla-bla> ([<bla-bla IP>]:17788) using "publickey".
```
But then
```
debug1: Sending command: nix-daemon --stdio
copying 0 paths...
starting pool of 15 threads
killing process 421682
kex_exchange_identification: read: Connection reset by peer
Connection reset by <bla-bla IP> port 22
```

rvem: Oh, looks like it fails somewhere further

src/push.rs
```diff
@@ -99,12 +99,58 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
...
+
+        match copy_command_status.code() {
+            Some(0) => (),
+            a => return Err(PushProfileError::CopyExit(a)),
+        };
+
+        // add remote host to building arguments too
+        build_command
+            .arg("--eval-store").arg("auto")
+            .arg("--store").arg(&store_address);
```
rvem, line 151:
```suggestion
            .arg("--store").arg(&store_address).env("NIX_SSHOPTS", ssh_opts_str.clone());
```

src/push.rs
rvem: FTR, I have the same issue even with NixOS -> NixOS deployment :see_no_evil:

hazelweakly: How I ended up "sorta solving this" myself was to take the path that deploy info spits out and then dead-reckoning run a little bash function with that path as the input.

```sh
function activate-on-host(){                                                            
  printf -v profile %q "$1"                                                          
  ssh user@FQDN "export PROFILE=$profile ; bash $profile/deploy-rs-activate" 
}                                                                                    
```

Which, indeed, simply runs the activation on the target

PhilTaken: I see, I did not encounter that error, very likely because I had it built locally before I tested remote building

the push_profile function does quite a lote at this point and most of it assumes the derivation is being built locally
splitting it up into a function to handle building locally and another to handle remote builds should resolve some of those difficulties

PhilTaken: besides, contrary to its name it also handles the building, not just the pushing :thinking:

PhilTaken:
@hazelweakly could you try again with the newest changes? 

make sure to gc before :wink:

src/push.rs
```diff
@@ -179,22 +146,77 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
...
+    let mut build_command = Command::new("nix");
+    build_command
+        .arg("build").arg(derivation_name)
+        .arg("--eval-store") .arg("auto")
+        .arg("--store").arg(&store_address)
+        .args(data.extra_build_args)
+        .env("NIX_SSHOPTS", ssh_opts_str.clone());
+
+    debug!("build command: {:?}", build_command);
+
```
rvem, line 195: Perhaps I'm missing something, but is profile activation performed within `build_profile_remotely`?

rvem: Oh, okay, profile activation is done within `deploy_profile`

src/push.rs
```diff
@@ -179,22 +146,77 @@ pub async fn push_profile(data: PushProfileData<'_>) -> Result<(), PushProfileEr
...
+    match copy_command_status.code() {
+        Some(0) => (),
+        a => return Err(PushProfileError::CopyExit(a)),
+    };
+
+    let mut build_command = Command::new("nix");
+    build_command
+        .arg("build").arg(derivation_name)
+        .arg("--eval-store") .arg("auto")
```
rvem, line 189: Tiny nitpick:
```suggestion
        .arg("--eval-store").arg("auto")
```

rvem (approved):
Tried both remote and non-remote builds, worked for me :tada:

hazelweakly:
@PhilTaken I can confirm that it works for me too (remotely). Thank you! This is very nice :)

PR #176, opened 2022-10-09
Add new activation strategy `boot` as equivalent to `nixos-rebuild boot`

Ma27:
This can be useful when e.g. deploying a kernel update to a target host. You usually plan a reboot (or kexec) after that to activate the new kernel. However you don't want to wait for services to be restarted first since these will be "restarted" anyways on the reboot. In cases like GitLab or the Atlassian stack this actually makes a difference.

This patch changes the following things:

* If `--boot` is provided, `nix-env -p profile-to-activate --set` is called for each deployed profile to make sure that it is activated automatically after a reboot.

* However, the actual activation (e.g. `switch-to-configuration switch`) is skipped. Instead:

  * For NixOS, `switch-to-configuration boot` is called to set the new profile as default in the bootloader.

  * For everything else, nothing else is done. The profile is already the new default (and thus picked up on the next boot).

cc @balsoft

### DIFF
--- a/src/bin/activate.rs
+++ b/src/bin/activate.rs
@@ -69,6 +69,10 @@ struct ActivateOpts {
     #[clap(long)]
     dry_activate: bool,
 
+    /// Don't activate, but update the boot loader to boot into the new profile
+    #[clap(long)]
+    boot: bool,
+
     /// Path for any temporary files that may be needed during activation
     #[clap(long)]
     temp_path: String,
@@ -363,6 +367,7 @@ pub async fn activate(
     confirm_timeout: u16,
     magic_rollback: bool,
     dry_activate: bool,
+    boot: bool,
 ) -> Result<(), ActivateError> {
     if !dry_activate {
         info!("Activating profile");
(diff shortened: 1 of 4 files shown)


Ma27:
cc @PhilTaken

src/bin/activate.rs
```diff
@@ -425,7 +431,7 @@ pub async fn activate(
             info!("Activation succeeded!");
         }
 
-        if magic_rollback {
+        if magic_rollback && !boot {
```
balsoft, line 434: I think there should be a warning saying that magic_rollback doesn't work with dry-activate or boot

balsoft (approved):
Eh, when I initially came up with the idea for deploy-rs, I wanted to avoid this coupling to profile activation scripts. But, given that (1) most people use deploy-rs for NixOS anyways, (2) this is quite useful, and (3) we have dry-activate already, I think it's ok.

Implementation LGTM

balsoft:
`nix flake check` passes

balsoft:
Thanks!

Issue #185, opened 2022-12-26
systemd-booted NixOS does not activate


notgne2:
It's the revenge of https://github.com/serokell/deploy-rs/issues/31

There is this offending code in nixpkgs - https://github.com/NixOS/nixpkgs/blob/3928cfa27d9925f9fbd1d211cf2549f723546a81/nixos/modules/system/boot/systemd/initrd.nix#L483-L490

The `-x` check will fail as in a real NixOS closure it is a file, but in ours it is actually a symlink to that file

(`-r-xr-xr-x     1 root root   4.5K Dec 31  1969 prepare-root` vs `lrwxrwxrwx     1 root root    103 Dec 31  1969 prepare-root -> /nix/store/8aicg4wzivp50jvshd8ix262vy00pnyf-nixos-system-peppaframe-23.05.20221222.652e92b/prepare-root`)

This results in seeing this message, and a system that will not function correctly (booting will complete, services will run, but `/run/current-system` etc will not get populated).

```
Dec 26 16:09:08 localhost initrd-nixos-activation-start[522]: /nix/store/lisipr5yinsb3xfbsid8fdrhgap8c3r0-activatable-nixos-system-peppaframe-23.05.20221222.652e92b does not look like a NixOS installation - not activating
```

There is no current workaround, besides not using `boot.initrd.systemd` (but that has other costs, such as no plymouth luks prompt). While this is currently avoidable, it sounds like this may become the default in the future, rendering deploy-rs completely unusable if this is not resolved.

notgne2:
Actually, I have no idea what's going on here... `-x` works fine it seems? It actually even goes as far as checking the symlink target to see if it's executable. That check should be working fine in theory, but something is failing in practice, or is it...

I actually did a quick test, the log message example above is real, and the path it mentions still exists, so I even gave it a quick test directly, and yup, that check won't fail and it should boot fine, although, err, well it definitely didn't.

notgne2:
Wondering now if it has something to do with conservative copying? I don't know how the initrd stuff works, although there exists a property `storePaths` listing store paths to copy, and I imagine there is a chance that the _real_ system profile isn't in it, and therefore the target of the symlink is genuinely missing at the point that the log-line occurred.

ElvishJerricco:
[...]

PR #188, opened 2023-01-08
Use more complex activation wrapper and add workaround for deploy-rs#185

notgne2:
There has to be a better way, but this should work around https://github.com/serokell/deploy-rs/issues/185 for now, I've tested it on my own system and I am able to successfully boot with systemd-initrd for the first time. This could have been a smaller patch but I also wanted to change the way the `activatable-` wrapping works to be be a little more sneaky and hopefully trip things up a little less.


notgne2:
To make it more discoverable, this would at least partially fix it upstream: https://github.com/NixOS/nixpkgs/pull/209829

imo we should do both (and also a patch to run the offending `test -x` in the sysroot chroot)

rvem:
@notgne2, may I ask you to squash and rebase your branch before merging?

PR #191, opened 2023-01-20
Set NixOS boot entry after activation instead of before

notgne2:
This solves an issue where your activation might fail, you reboot, and then you are back on the failed activation :(


notgne2:
Note: the danger of this is that when rolling back, if the activation of the rollback fails then it will not roll back your boot entry

PR #203, opened 2023-04-18
[#202] Provide '^out' suffix for deriver on newer nix

### LINKED ISSUE #202: Deploy-rs errors with `nix 2.15.0`: The interpretation of store paths arguments ending in `.drv` recently changed
The latest `pkgs.nixUnstable` release of nix, version `2.15.0`, breaks `deploy-rs`. The same configuration with nix `2.14.1` works correctly. This seems to have caused #199 and why `remoteBuilder = true` fixes this error, likely due to differing nix versions between the local and remote machine.

The logs after a weekly `nix flake update` and `deploy .#host` command:
```
% deploy .#nona --debug-logs
🚀 ❓ [deploy] [DEBUG] Checking for flake support
🚀 ℹ️ [deploy] [INFO] Running checks for flake in .
🚀 ℹ️ [deploy] [INFO] Evaluating flake in .
🚀 ℹ️ [deploy] [INFO] The following profiles are going to be deployed:
[nona.system]
user = "root"
ssh_user = "sofi"
path = "/nix/store/iignc95ivrjl2xisb4d57d6avhka9qhy-activatable-nixos-system-nona-23.05.20230417.4aa62d5"
hostname = "nona.hosts.byte.surf"
ssh_opts = []

🚀 ❓ [deploy] [DEBUG] Finding the deriver of store path for /nix/store/iignc95ivrjl2xisb4d57d6avhka9qhy-activatable-nixos-system-nona-23.05.20230417.4aa62d5
🚀 ℹ️ [deploy] [INFO] Building profile `system` for node `nona`
warning: The interpretation of store paths arguments ending in `.drv` recently changed. If this command is now failing try again with '/nix/store/7k21mmwi0mcfwqpkr9zjm1z7191a7qg0-activatable-nixos-system-nona-23.05.20230417.4aa62d5.drv!*'
🚀 ❌ [deploy] [ERROR] Failed to push profile: Activation script deploy-rs-activate does not exist in profile.
Did you forget to use deploy-rs#lib.<...>.activate.<...> on your profile path?
```

rvem:
Problem: Since 2.15 nix no longer reference '.drv' as derivation
outputs. At the same time, nix before '2.13' doesn't support '.drv'
special suffix handling.

Solution: Provide '^out' suffix for the profile deriver in case
'nix path-info <...>.drv' returns the same '<...>.drv' path.
In other cases either an error about the build result not being present
in the /nix/store is returned or an actual build result path is
returned.

Resolves #202


dschrempf:
I can confirm that deployment works with this branch and Nix 2.15.

imsofi:
I can also confirm this works correctly. However, it will not use substitutes that you have set up yourself in `extra-substituters`. So you will be building everything that would otherwise have been substituted.

Using `nix build .#nixosConfigurations.nona.config.system.build.toplevel` will pull from my substituter, so this is an issue of how deploy-rs is pulling from the cache with `nix 2.15.0`.

rvem:
Hmm, that's unexpected, I'll take a look on why these derivations diverge

rvem:
> so this is an issue of how deploy-rs is pulling from the cache with `nix 2.15.0`.

I believe that nothing has really changed on the `deploy-rs` side, `deploy.nodes.<node>.profiles.system.path` uses the corresponding `nixosConfiguration` as input and thus it should be fetched from cache :thinking:

imsofi:
An interesting point I noticed was that after using `nix build` to fetch from my personal cache, the deploy command would happily use the derivations that had been downloaded.

I will try to see if the derivations change if I let deploy-rs build it.

imsofi:
Just got some time. Both derivations end up as the same hash. So there is something about `deploy-rs` where it does not realize that there exists a cache in `nixConfig.extra-substituters` from the  `flake.nix` file with that package when using `nix 2.15.0` and the current PR here.

```
# nix build - was downloaded from cache
/nix/store/2lyxfjv87kvcns16rs439n5idak5p71c-jellyfin-10.8.8
/nix/store/24ml78h7i7h7a384bhnf00ansakq4zvv-jellyfin-web-10.8.8


# deploy-rs - was built from source
/nix/store/2lyxfjv87kvcns16rs439n5idak5p71c-jellyfin-10.8.8
/nix/store/24ml78h7i7h7a384bhnf00ansakq4zvv-jellyfin-web-10.8.8
```

rvem:
Can you check if this behaviour also happens for nix before `2.15.0` and `deploy-rs` before this PR?

imsofi:
You are indeed correct. This actually happens on `nix 2.14.1` as well. Now i am unsure how i have not noticed this before.

dschrempf:
That's funny because I just have installed the same package on my server with deploy-rs and it definitely fetched the package from cache.

imsofi:
@dschrempf Yeah i am running my own custom patchset of jellyfin, which is why i am using it as an example. I should have specified that it was not the `nixpkgs` jellyfin.

imsofi (approved):
The issues I described previously ended up being an important but unrelated issue (#204), this PR sets out solving deploy-rs incompatibility with nix 2.15+ while also staying backwards compatible.

src/push.rs
```diff
@@ -215,6 +223,16 @@ pub async fn build_profile(data: PushProfileData<'_>) -> Result<(), PushProfileE
         &data.deploy_data.profile.profile_settings.path
     );
 
+    let nix_version_output = Command::new("nix").arg("--version")
+        .output()
+        .await.map_err(PushProfileError::NixVersion)?;
+
+    let version_regex = Regex::new(r"(\d+\.\d+\..+)").unwrap();
+    let nix_version = version_regex.
```
PhilTaken, line 231: "idiomatic" rust code keeps the `.` at the beginning of the next line as in

```
let nix_version = version_regex
    .find(std::str::from_utf8(&nix_version_output.stdout)
        .map_err(PushProfileError::NixVersionUtf8)?)
    .map(|v : regex::Match| -> &str { v.as_str() })
    .ok_or(PushProfileError::NixVersionParse)?;
```

maybe we should introduce some pre-commit hooks for deploy-rs or at least a formatter via flakes, those could also be integrated into the ci

src/push.rs
```diff
@@ -215,6 +223,16 @@ pub async fn build_profile(data: PushProfileData<'_>) -> Result<(), PushProfileE
         &data.deploy_data.profile.profile_settings.path
     );
 
+    let nix_version_output = Command::new("nix").arg("--version")
+        .output()
+        .await.map_err(PushProfileError::NixVersion)?;
```
PhilTaken, line 228: stylistically, I would put the `await` either on it's own line or behind `output()`

src/push.rs
```diff
@@ -215,6 +223,16 @@ pub async fn build_profile(data: PushProfileData<'_>) -> Result<(), PushProfileE
...
+
+    let version_regex = Regex::new(r"(\d+\.\d+\..+)").unwrap();
```
PhilTaken, line 230: I'm always a bet hesitant about regexes, this one assumes that the nix version output will always adhere to the `major.minor` format and fails if it doesn't :/

rvem: Fair enough, do you have any ideas how can this be made more robust?

rvem: Perhaps, we can try `^out` first and then fallback to the old format in case the first one fails

rvem: Also, `nix path-info` might do the thing, but it'll require to parse output either stdout or stderr in case the resulting path doesn't exist yet

rvem: Implemented an approach that uses `nix path-info`, which seems to be working and I believe it's a bit more robust than version grepping

rvem:
@imsofi, I slightly changed the approach, can you check if the new version still works as expected for you?

imsofi:
Seems to work fine :+1:

Had a one off `error: failed to start SSH connection to 'user@hostname'` that i have never seen before. But rerunning the command worked correctly. Unsure if that indicates anything or if my server just decided to be wonky this one time.

src/push.rs
PhilTaken: I would agree that `path-info` is probably much more robust

PhilTaken (approved):

PhilTaken:
> error: failed to start SSH connection to 'user@hostname'

that doesn't seem to be related to this pr (fingers crossed) but seems worth investigating if the issue can be reproduced

imsofi:
Never seen it since. So it's likely just a badly placed connection issue just as I tested the new version of the PR 😅

PR #205, opened 2023-04-19
[Chore] Handle 'temp_path' as an actual 'Path' instead of 'String'

rvem:
Problem: 'temp_path' and 'lock_path' are handled as 'String'. This can be a problem when the 'temp_path' directory is a symlink on the target system, e.g. this is the case with the default '/tmp' and macOS, where this directory is actually a symlink to '/private/tmp'.

Solution: Handle 'temp_path' and 'lock_path' as actual Paths. This should help to avoid canary file path mismatches when checking filesystem events.

As a side effect, also update the 'notify' dependency to the latest stable version.


srid (approved):
Not reviewing the code per se, but I've tested this PR to work on a M1 mac -> M1 mac deployment, thanks @rvem 

https://github.com/nammayatri/darwin-ci/pull/1

PhilTaken (approved):
This is way better than before. Code looks very clean to me and is better to reason about than always pretending a string represents a path.

Issue #204, opened 2023-04-19
`deploy` command does not pull in flake attributes on build.


imsofi:
When running `deploy .#hostname`, deploy-rs will use the `nix show-derivation` command to find the actual `.drv` path of the NixOS configuration, and then later use it for the nix build step, for example `nix build /nix/store/[...]-nixos-system-[hostname]-[...].drv`.

https://github.com/serokell/deploy-rs/blob/8c9ea9605eed20528bf60fae35a2b613b901fd77/src/push.rs#L241-L244

https://github.com/serokell/deploy-rs/blob/8c9ea9605eed20528bf60fae35a2b613b901fd77/src/push.rs#L71-L73

 However, running `nix build` in this way will disconnect the build step from the `flake.nix`. This is problematic since it makes `nixConfig` configuration not apply when building trough deploy-rs, and this argument is often used to define `extra-substituters` per repository.
 
 I propose that deploy-rs will use `.#nixosConfigurations.[hostname].config.system.build.toplevel` instead as the argument for `nix build` to let it be flake aware.

This was discovered during debugging odd behavior in the latest nix 2.15.0 release in #203.

CC: @rvem

imsofi:
![image](https://user-images.githubusercontent.com/20756843/233054515-9720a09f-1084-4b22-bac6-af8e06975738.png)

Actually, a really simple method to solve this is using `/path/to/flake.nix#deploy.nodes.[hostname].profile.system.path` instead of the `.drv` path from the `.path` directly.

imsofi:
![image](https://user-images.githubusercontent.com/20756843/233060731-1e5b40c2-73df-4654-9612-fc82aa897016.png)

Got a MVP working of what i described above with `flake.path`.

However i think using `--option` and adding `nixConfig` to `data` would be best. As using `flake.path` would only work locally and not remotely.

imsofi:
From closer looking, there is no great solution for this. As an auto-accepting `--option` may be dangerous, as `nixConfig` can be partially applied by each user, or completely denied.

`nixos-rebuild` uses `--builders` for remote builders, which will use the `nixConfig` on the local machine and not touch the builder unless needed. Deploy-rs uses ssh to run `nix build` directly. While we could copy the `flake.nix` instead on the remote, but it would not use the local `nixConfig` list. So either an interactive mode to say yes/no to the options, or an automatic yes trough the `--accept-flake-config` argument would be needed here.

[...]

PR #206, opened 2023-04-19
fix: let nixConfig be used at build step

imsofi:
Allows `nix build` to become flake aware. This is important to allow nixConfig based changes to be used at build time. For example when setting `extra-substituters` through the flake.nix file.

Partially fixes: #204, this PR does not fix the remote builder.

### DIFF
--- a/src/push.rs
+++ b/src/push.rs
@@ -69,7 +69,11 @@ pub async fn build_profile_locally(data: &PushProfileData<'_>, derivation_name: 
     };
 
     if data.supports_flakes {
-        build_command.arg("build").arg(derivation_name)
+        // Use flake based path over derivation_name to let nixConfig options be used for build
+        build_command.arg("build").arg(format!(
+            "{}#deploy.nodes.{}.profiles.{}.path",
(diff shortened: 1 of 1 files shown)


balsoft (changes requested):
This evaluates the derivation twice. This will result in quite a massive slow-down for many deployments.

We should either fix Nix evaluation cache to handle this properly, or not evaluate the derivation path when first evaluating the `deploy` output.

imsofi:
That is some good spotting! I had not thought about that.

Handling `nixConfig` has ended up being much more of a pain than first thought. The only other idea that i have is running something similar to `nix eval --json --file flake.nix nixConfig` at the data step, using it to populate a `nix-config` list that could be applied with `--option`. This should not evaluate anything under the flake inputs and flake outputs for this second pass.

Originally wanted to avoid this as re-implementing how `nix build` works natively i feel could have small unexpected changes that could be hard to reason and debug, as well as there not being an easy way to handle the per permission toggle that `nixConfig` follows.

imsofi:
Dropping this PR due to my Rust knowledge not being up for the job. If anyone else wants to pick this up, feel free to.

PR #207, opened 2023-04-19
Make it possible to not rebuild deploy-rs

### LINKED ISSUE #163: Have an officially hosted and/or endorsed binary cache for deploy-rs
Provide a binary cache for deploy-rs, for example trough the use of cachix. I find it strange that everyone using deploy-rs still needs to compile it on their own, or set up their own binary cache for the use with their teams.

There used to be a personal cachix repo for deploy-rs, but it is now unmaintained https://github.com/cmacrae/deploy-rs-binary-builds

@cmacrae do you have any potential advice for creating an official cache?

Flakebi:
Use the deploy-rs from the final packages set. This can avoid rebuilding deploy-rs when using it in a nixos config. It can use the version cached in nixpkgs.

Also add instructions to the readme on how to craft an overlay that uses nixpkgs deploy-rs.

This should fix #163.

### DIFF
--- a/README.md
+++ b/README.md
@@ -86,6 +86,33 @@ A basic example of a flake that works with `deploy-rs` and deploys a simple NixO
 }
 ```
 
+In the above configuration, `deploy-rs` is built from the flake, not from nixpkgs. To take advantage of the nixpkgs binary cache, the deploy-rs package can be overwritten in an overlay:
+
+```nix
+{
+  # ...
+  outputs = { self, nixpkgs, deploy-rs }: let
+    system = "x86_64-linux";
+    # Unmodified nixpkgs
+    pkgs = import nixpkgs { inherit system; };
+    # nixpkgs with deploy-rs overlay but force the nixpkgs package
+    deployPkgs = import nixpkgs {
+      inherit system;
+      overlays = [
+        deploy-rs.overlay
+        (self: super: { deploy-rs = { inherit (pkgs) deploy-rs; lib = super.deploy-rs.lib; }; })
+      ];
+    };
+  in {
+    # ...
+    deploy.nodes.some-random-system.profiles.system = {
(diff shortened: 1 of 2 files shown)


rvem (approved):

rvem:
Thanks!

Stunkymonkey:
I am seeing the `checks` are disabled. Is this intentional? Or does it not make sense?

rvem:
> Is this intentional?

Hmm, not really, not sure why checks weren't triggered for external contributor PR

wmertens:
@Flakebi question, how does this overlay work? So there's the overlay from the flake and then there's this expression:

```nix
  (self: super: { deploy-rs = { inherit (unstablePkgs) deploy-rs; lib = super.deploy-rs.lib; }; })
```

the way I read the spec is that this function should return an attrset of packages, however it's returning `{deploy-rs; lib}` as a "package". Is this an undocumented feature of overlays?

Flakebi:
~~There is nothing undocumented~~ There is no hidden feature of overlays here.
The “package”, which is actually a set, not a derivation, as you noticed, is the structure defined in this flake.
The overlay defines the same set as this flake, keeping the `lib` attribute intact and overriding the `deploy-rs` attribute, which is the actual package.

This structure is defined here in the flake:
https://github.com/serokell/deploy-rs/blob/724463b5a94daa810abfc64a4f87faef4e00f984/flake.nix#L31-L42

wmertens:
Thank you for the clarification. That's mighty confusing 😅

teutat3s:
Thank you for this!

Is there a way to make this work when deploy host `system` != target host `system`? Currently the overlay is hard-coded to `x86_64-linux` and that causes an "exec format error" when trying to deploy to a `aarch64-linux` host.

PR #209, opened 2023-04-28
Add riscv64-linux to supported systems

0x4A6F:


trueNAHO (commented):
The currently pending commit https://github.com/serokell/deploy-rs/pull/345/commits/428b3a54cabd0fd93fa487979e225c43de679785 ("Add systems input to leverage extensible flake systems pattern") from https://github.com/serokell/deploy-rs/pull/345 allows externally overriding the list of supported systems.

PR #220, opened 2023-06-15
Add extra-ssh-opts flag

mikroskeem:
This is useful in situations when there's a need to pass on extra ssh client options, e.g. in CI pipeline to set different hostname, port (VPN requiring different hostnames) without overriding options set in deploy configuration.


mikroskeem:
Bump, is there any interest in this feature?

PR #231, opened 2023-09-06
[#201] Deduce profile directory during activation

### LINKED ISSUE #201: Update default profiles location
Since https://github.com/NixOS/nix/pull/5226 user profiles are by default located in `$XDG_DATA_DIR/nix/`profiles`.
However, the profiles directory is currently [hardcoded](https://github.com/serokell/deploy-rs/blob/2a3c5f70eee04a465aa534d8bd4fcc9bb3c4a8ce/src/lib.rs#L371) in `deploy-rs`, thus `deploy-rs` will likely break for newer nix.

Note that `deploy-rs` should probably handle both old and new profiles directories to avoid introducing breaking changes for users.

rvem:
Problem: Since https://github.com/NixOS/nix/pull/5226 nix profiles for users are stored in 'XDG_STATE_HOME' or 'HOME' directory. However, 'deploy-rs' still expects profiles to be present in '/nix/var/nix/profiles/per-user'. As a result, an attempt to deploy a profile with newer nix may fail with an error about non-existing files.

Solution: Instead of deducing the profile path prior to ssh'ing and actual activation, deduce the path to the profile as a part of 'activate-rs' invocation.

Now if the profile path is not specified explicitly as an attribute in profile within the deploy flake, the path to the profile is determined based on the user to which the profile belongs and on the values of 'XDG_STATE_HOME' and 'HOME' variables.
Additionally, if the old profile directory (in
'/nix/var/nix/profiles/per-user') for a given user already exists, it is used instead for the sake of backward compatibility.

Resolves #201


src/deploy.rs
```diff
@@ -157,7 +174,7 @@ fn test_wait_command_builder() {
 struct RevokeCommandData<'a> {
     sudo: &'a Option<String>,
     closure: &'a str,
-    profile_path: &'a str,
```
PhilTaken, line 160: Love expressive types, glad this is gone

PhilTaken (approved):
lgtm

rvem:
@PhilTaken, I updated `root` user profiles handling a bit to match the documentation, could you please re-review?

Sereja313 (approved):
i didnt notice anything suspicious

src/bin/activate.rs
```diff
@@ -459,6 +480,57 @@ async fn revoke(profile_path: String) -> Result<(), DeactivateError> {
...
+                _ => {
+                    let old_user_profiles_dir =
+                        format!("{}/profiles/per-user/{}", nix_state_dir, profile_user);
+                    // To stay backward compatible
+                    if Path::new(&old_user_profiles_dir).exists() {
+                        Ok(format!("{}/{}", old_user_profiles_dir, profile_name))
+                    } else {
+                        // https://github.com/NixOS/nix/blob/2.17.0/src/libstore/profiles.cc#L308
+                        let state_dir = env::var("XDG_STATE_HOME").or_else(|_| {
+                            dirs::home_dir()
```
PhilTaken, line 519: why not just use `dirs::state_dir()`?

PhilTaken: also, all of the Paths in this function are handled as strings, I think explicit Path types would be preferable

rvem: > why not just use `dirs::state_dir()`?

Nice catch. However, AFAICS, it will return `None` on macOS, while `nix` will fallback to `$HOME/.local/state` on both Linux and macOS.

rvem: Another possible issue with macOS is that `sudo` there doesn't change the `HOME` directory :confused: 
So non-root profiles on macOS deployment sounds like a huge PITA

PhilTaken: > Nice catch. However, AFAICS, it will return None on macOS, while nix will fallback to $HOME/.local/state on both Linux and macOS. I think it's better to mock the nix behaviour here and avoid using state_dir

good point, lgtm then :)

PR #234, opened 2023-09-17
Replace jsonschema-cli with check-jsonschema

fd:
`jsonschema-cli` is deprecated and will be removed in the future. The recommended replacement is `check-jsonschema`.

I tested the changes with my setup, the new cli gives a positive validation result on stdout.


rvem (approved):

rvem:
Thanks!

PR #236, opened 2023-09-18
[Chore] Run CI checks on 'pull_request'

rvem:
Problem: We want to be able to run CI checks on PRs from external forks.
However, this is only possible with 'on: pull_request', while currently
CI is triggered 'on: push'

Solution: Change CI triggering condition to 'on: pull_request'.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -1,5 +1,5 @@
 name: Nix flake check
-on: push
+on: pull_request
 
 jobs:
   check:


Sereja313 (approved):

PhilTaken:
Is this not security relevant? Docker has a bunch of problems that can lead to escaping the container. Malicious PRs on selfhosted runners (if they still are for the serokell org) could exploit those vulnerabilities if not properly secured.

rvem:
> Is this not security relevant?

External jobs require manual approval in order to run, so this should be relatively safe as long as we check what we actually run on our CI :sweat_smile:

PR #239, opened 2023-10-01
Add a flag for building configuration with nix-output-monitor

stuebinm:
I like using the [nix-output-monitor](https://github.com/maralorn/nix-output-monitor) for my builds sometimes, so I added a `--nom` flag to `deploy` to make it call `nom` instead of `nix` when building the system configuration.

This is a pretty bare-bones change (e.g. it does not check if `nom` is actually present on the `$PATH`, though this could be added), and I'm not sure if you're interested in such a feature for `deploy-rs` in the first place — but I thought I'd leave this PR here in case it's useful to you.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -100,6 +100,9 @@ pub struct Opts {
     /// Which sudo command to use. Must accept at least two arguments: user name to execute commands as and the rest is the command to execute
     #[clap(long)]
     sudo: Option<String>,
+    /// Call nom (nix-output-monitor) when building the configuration.
+    #[clap(long)]
+    nom: bool,
 }
 
 /// Returns if the available Nix installation supports flakes
@@ -418,6 +421,7 @@ async fn run_deploy(
     boot: bool,
     log_dir: &Option<String>,
     rollback_succeeded: bool,
+    use_nom: bool,
 ) -> Result<(), RunDeployError> {
     let to_deploy: ToDeploy = deploy_flakes
         .iter()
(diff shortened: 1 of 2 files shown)


notgne2:
I like to run just `deploy` personally, I think others may feel the same - maybe it would make sense to have this configurable from the flake too?

Not sure if others would agree with merging this in, as I guess it could seem like feature creep. I've never seen nix-output-monitor before, but it looks neat, and I'd like to, so personally I wouldn't mind a way to use it. Are there any easy alternatives to adding integration of it, such as wrapping `deploy` with something that replaces the `nix` binaries with `nom` binaries?

rvem:
I agree that
>  Are there any easy alternatives to adding integration of it, such as wrapping `deploy` with something that replaces the `nix` binaries with `nom` binaries?

Would be easier and less error-prone

stuebinm:
> such as wrapping deploy with something that replaces the nix binaries with nom binaries?

I considered that when implementing it, but it seems hard — `deploy` does a lot of calls to `nix`, and not all make sense to pipe through `nom` in my opinion (e.g. it instantiates the flake to get information about the target first, where nix-output-manager would just clutter the terminal, it [uses `nix eval` to check if the `nix` present supports flakes](https://github.com/serokell/deploy-rs/blob/master/src/cli.rs#L106) but `nom eval` doesn't even exist, …), so I'm not sure if there's a way to do this without having `deploy` be aware of it — at least I couldn't think of any.

> maybe it would make sense to have this configurable from the flake too?

Oh that could be nice, yeah; I didn't think of that (I already always run `deploy -ks` out of habit, so the extra flag didn't bother me much).

Maybe I should've opened a more general discussion issue first :see_no_evil:

PR #246, opened 2023-12-08
[#245] Return non-zero exit code in case of confirmation timeout

### LINKED ISSUE #245: question: how to ensure non-0 exit code of rolled back deployment?
I'm using deploy-rs personally and at work, and this has been a longstanding issue for a long time, that we falsly believe that we have new deployments in production, only to expand the CI pipeline logs to see that the deployment didn't success, but "successfully" rolled back. This is giving us bad signals about the state of our system. Is there a command line flag or something that I'm missing out on, that ensures that any non-passing deployment is always, under all circumstances, returning non-0 exit code?

rvem:
Problem: When profile activation confirmation fails due to
confirmation timeout and performs a rollback, zero exit code is
returned. Such a behavior is confusing since rollback usually means
something went wrong during deployment and it shouldn't return
successful exit code.

Solution: Explicitly return confirmation waiting error instead of
printing it and silently signaling success.

Resolves #245

### DIFF
--- a/src/bin/activate.rs
+++ b/src/bin/activate.rs
@@ -229,6 +229,8 @@ pub enum ActivationConfirmationError {
     CreateConfirmFile(std::io::Error),
     #[error("Could not watch for activation sentinel: {0}")]
     Watcher(#[from] notify::Error),
+    #[error("Error waiting for confirmation event: {0}")]
+    WaitingError(#[from] DangerZoneError),
 }
 
 #[derive(Error, Debug)]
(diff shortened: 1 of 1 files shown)


rvem:
> Upon closer look it appears that this change is limited to the remote side, the deploying host should receive the timeout reliably 🤔

AFAICS, the corresponding `activate` is invoked via `ssh <target> activate-rs <options>`, so `activate` is indeed run on the remote side but on the foreground of the ssh session

PhilTaken:
right, and then the exit code is propagated through the connection. perfect then.

LGTM

PhilTaken (approved):

PR #257, opened 2024-02-03
Add support for entering sudo password interactively

### LINKED ISSUE #78: Password based sudo
After bootstrapping a host with regular nixos-install and setting a password for `nixos` user, I then tried to deploy subsequent generations with deploy-rs. However, even if I turned on interactive mode, I was not able to enter the sudo password.

Either I didn't understand how to do it (ux problem) or it is plain not possible at the moment (i think this is the case).

Some people might be reluctant to configure passwordless sudo as base line, hence a way is needed to supply elevation credentials during deployment.

Coming from https://github.com/divnix/devos/pull/197#issuecomment-801636979

n-hass:
This adds a prompt for entering the sudo password interactively, so that system deployments can be activated when using non-root sshUsers. Fixes #78.

I have chosen to handle this locally (in deploy) and pipe the password to stdin of the ssh session, rather than via psuedo-terminal as this is more reliable. This also means that magicRollback works, unlike the "-t" workaround.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -103,6 +103,9 @@ pub struct Opts {
     /// Which sudo command to use. Must accept at least two arguments: user name to execute commands as and the rest is the command to execute
     #[clap(long)]
     sudo: Option<String>,
+    /// Prompt for sudo password during activation.
+    #[clap(long)]
+    interactive_sudo: Option<bool>,
 }
 
 /// Returns if the available Nix installation supports flakes
@@ -538,7 +541,25 @@ async fn run_deploy(
             log_dir.as_deref(),
         );
 
-        let deploy_defs = deploy_data.defs()?;
+        let mut deploy_defs = deploy_data.defs()?;
+
+        if deploy_data.merged_settings.interactive_sudo.unwrap_or(false) {
+            warn!("Interactive sudo is enabled! Using a sudo password is less secure than correctly configured SSH keys.\nPlease use keys in production environments.");
+            
+            if deploy_data.merged_settings.sudo.is_some() {
+                warn!("Custom sudo commands should be configured to accept password input from stdin when using the 'interactive sudo' option. Deployment may fail if the custom command ignores stdin.");
+            } else {
+                // this configures sudo to hide the password prompt and accept input from stdin
+                // at the time of writing, deploy_defs.sudo defaults to 'sudo -u root' when using user=root and sshUser as non-root
+                let original = deploy_defs.sudo.unwrap_or("sudo".to_string());
+                deploy_defs.sudo = Some(format!("{} -S -p \"\"", original));
+            }
+
+            info!("You will now be prompted for the sudo password for {}.", node.node_settings.hostname);
+            let sudo_password = rpassword::prompt_password(format!("(sudo for {}) Password: ", node.node_settings.hostname)).unwrap_or("".to_string());
+
+            deploy_defs.sudo_password = Some(sudo_password);
+        }
 
         parts.push((deploy_flake, deploy_data, deploy_defs));
     }
@@ -665,6 +686,7 @@ pub async fn run(args: Option<&ArgMatches>) -> Result<(), RunError> {
         dry_activate: opts.dry_activate,
         remote_build: opts.remote_build,
         sudo: opts.sudo,
+        interactive_sudo: opts.interactive_sudo
     };
 
     let supports_flakes = test_flake_support().await.map_err(RunError::FlakeTest)?;
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -139,6 +139,7 @@ dependencies = [
  "merge",
  "notify",
  "rnix",
+ "rpassword",
  "serde",
  "serde_json",
  "signal-hook",
@@ -667,6 +668,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "rpassword"
+version = "7.3.1"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "80472be3c897911d0137b2d2b9055faf6eeac5b14e324073d83bc17b191d7e3f"
+dependencies = [
+ "libc",
+ "rtoolbox",
+ "windows-sys 0.48.0",
+]
+
+[[package]]
+name = "rtoolbox"
+version = "0.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c247d24e63230cdb56463ae328478bd5eac8b8faa8c69461a77e8e323afac90e"
+dependencies = [
+ "libc",
+ "windows-sys 0.48.0",
+]
+
+[[package]]
 name = "rustc-hash"
 version = "1.1.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -859,6 +881,7 @@ dependencies = [
  "autocfg",
  "bytes",
  "libc",
+ "memchr",
  "mio 0.7.6",
  "num_cpus",
  "once_cell",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -24,7 +24,7 @@ serde = { version = "1.0.104", features = [ "derive" ] }
 serde_json = "1.0.48"
 signal-hook = "0.3"
 thiserror = "1.0"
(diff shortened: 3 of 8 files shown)


n-hass:
Can I also make a suggestion that some refactoring take place? The same pattern is duplicated like ~5-6 times to build a process::Command with the ssh options etc. I think a factory-style function would be helpful here and would make the code significantly more readable.

I'm happy to do this in this PR or make another.

PhilTaken:
@n-hass 

Thanks for your contribution!

I'm not sure how I feel about transmitting passwords for sudo. Generally, you should use ssh keys with attached permissions instead.

As for the refactoring, feel free to open another PR since it is not related to this one.

n-hass:
Agreed, keys are always safer. However, from a security perspective, I believe this technique is sound.

Functionally this is no different than SSH'ing as the non-root user and entering the sudo password interactively through the tty, as deploy-rs would pipe the password through to the session's stdin after the session has been opened. The password is never transmitted or stored plaintext (apart from when in-memory in deploy-rs).

I think some, myself included, are more comfortable with this than widening SSH permissions for root, or using a sudo pam / ssh-agent forwarding workaround

n-hass:
Sorry, forgot to tag @PhilTaken . See above

src/cli.rs
```diff
@@ -103,6 +103,9 @@ pub struct Opts {
     /// Which sudo command to use. Must accept at least two arguments: user name to execute commands as and the rest is the command to execute
     #[clap(long)]
     sudo: Option<String>,
+    /// Prompt for sudo password during activation.
+    #[clap(long)]
+    interactive_sudo: Option<bool>,
```
rvem, line 108: May I ask you to also update https://github.com/serokell/deploy-rs/blob/master/interface.json to reflect the new option?

rvem (commented):
First of all, thanks for your contribution:)

> Perhaps a warning/'disclaimer' could be included somewhere to encourage use of keys in a production environment?

I agree, it'd be nice to print a warning that interactive sudo is used.

Additionally, I have a single comment regarding the JSON schema update that is affected by your PR

src/cli.rs
n-hass: Sure, pushing this now. I also noticed some other options may be missing from the schema, including `remoteBuild` and `sudo`.

src/cli.rs
```diff
@@ -544,14 +544,16 @@ async fn run_deploy(
...
             let sudo_password = rpassword::prompt_password(format!("(sudo for {}) Password: ", node.node_settings.hostname)).unwrap_or("".to_string());
 
             deploy_defs.sudo_password = Some(sudo_password);
-        
+            
+            // this instructs sudo to hide the password prompt and accept input from stdin
             if let Some(original) = deploy_defs.sudo.clone()  {
                 if !original.contains("-S") {
-                    deploy_defs.sudo = Some(format!("{} -S -p \"\"", original));
+                    deploy_defs.sudo = Some(format!("{} -S -p \"\"", original));
```
rvem, line 556: This piece of diff looks suspicious :sweat_smile:

src/cli.rs
rvem: Indeed, good catch

src/cli.rs
rvem: An additional concern I have regarding this line is that alternative some alternative `sudo` commands may not support provided flags, e.g. `doas`. I'm not sure what is the best way to fix this, perhaps, if we use custom `sudo` attribute together with `interactiveSudo = true;` we should require updating `sudo` attribute provide required flags instead of hardcoding them here

n-hass: Yeah I thought this may be too bold to assume...

Do you think maybe the thing to do is:
If no custom `sudo` attribute is defined, leave this code as-is. If the user has defined a custom `sudo` attribute, print a message along the lines of "ensure your sudo command can accept the password from stdin"? Or if theres already too many messages, wait until deployment fails and print a message like "you're using a custom `sudo` and `interactiveSudo`. Ensure your custom sudo can accept the password from stdin for this to work."

rvem: Yes, such an approach sounds good to me

rvem: I think it's better to print the warning before, because the deployment may fail due to a different error and the message after might potentially be misleading

src/cli.rs
```diff
@@ -545,19 +545,24 @@ async fn run_deploy(
...
         if deploy_data.merged_settings.interactive_sudo.unwrap_or(false) {
             warn!("Interactive sudo is enabled! Using a sudo password is less secure than correctly configured SSH keys.\nPlease use keys in production environments.");
+            
+            if deploy_data.merged_settings.sudo.is_some() {
+                warn!("Custom sudo commands should be configured to accept password input from stdin when using the 'interactive sudo' option. Deployment may fail if the custom command ignores stdin.");
+            } else {
+                // this configures sudo to hide the password prompt and accept input from stdin
+                if let Some(original) = deploy_defs.sudo.clone() {
+                    if !original.contains("-S") {
+                        deploy_defs.sudo = Some(format!("{} -S -p \"\"", original));
```
rvem, line 555: This line implicitly implies that the provided `deploy_defs.sudo` supports both `-S` and `-p` options. IMO, it's better to leave all options to be provided as a custom `sudo` instead of changing what is provided by the flake/command line option, i.e. completely avoid modifying `deploy_defs.sudo` if it's not a `None`
What do you think?

src/cli.rs
```diff
@@ -545,19 +545,24 @@ async fn run_deploy(
 
         if deploy_data.merged_settings.interactive_sudo.unwrap_or(false) {
             warn!("Interactive sudo is enabled! Using a sudo password is less secure than correctly configured SSH keys.\nPlease use keys in production environments.");
+            
+            if deploy_data.merged_settings.sudo.is_some() {
+                warn!("Custom sudo commands should be configured to accept password input from stdin when using the 'interactive sudo' option. Deployment may fail if the custom command ignores stdin.");
+            } else {
+                // this configures sudo to hide the password prompt and accept input from stdin
+                if let Some(original) = deploy_defs.sudo.clone() {
```
rvem, line 553: Hmm, I'm not sure if this `if` is needed now. IIUC, `deploy_data.merged_settings.sudo` is `None` at this point and we just simply need to append `-S -p ""` options to it. Am I missing something?

n-hass: When I was testing, given the following node config:
```
nodename = {
        hostname = "1.2.3.4";
        # Use the dynamically derived port number
        sudo = "doas";
        interactiveSudo = true;
        profiles.system = {
          sshUser = "deploy";
          user = "root";
          path = deploy-rs.lib.aarch64-linux.activate.nixos nixosConfigs.nodename;
        };
      };
      ```
      
`deploy_data.merged_settings.sudo` at this point is "doas". if I leave that custom sudo config out of the flake, ie dont define a sudo option,`deploy_data.merged_settings.sudo` is None at the time of this condition. is that intended? If so, it seems like a good way to differentiate when it is safe to automatically append `-S -p ""`, and when it should be left to the user to configure. Happy to do it another way if you suggest

Comment (n-hass) [reply src/cli.rs:553]: Ah. Sorry, I understand now - the innermost `if` should not be needed, but I noticed that when sudo overrides are passed via the cli ie like `--sudo='doas'`, then they do not get merged into the deploy_data.merged_settings. I was going to open another PR as I believe that could be a bug. This innermost `if`, however, covers the case where `deploy_defs` could somehow already have '-S'  in it.

**Review by rvem — APPROVED** [review]

**Comment by n-hass** [comment]
Thanks @rvem! Is it okay if I squash?

**Comment by rvem** [comment]
Yeah, sure

**Comment by rvem** [comment]
Thanks!

**Comment by munnik** [comment]
Hi,

After reading above comments and other documentation, I still don't understand why this message appears when doing a deployment:
```
🚀 ⚠️ [deploy] [WARN] Interactive sudo is enabled! Using a sudo password is less secure than correctly configured SSH keys.
```
I use interactive sudo, but why is it less secure than _correctly configured SSH keys_? For login, I use SSH keys, do I need to set other SSH keys? The only documentation I see is to either allow root to log in with SSH keys or to allow sudo without password. I think both are less secure than needing both an SSH key to login and a password to become root. Am I missing something?

PR #278, opened 2024-06-11
add meta.mainProgram to rust drv

emanueljg:
Most importantly, this fixes `lib.getExe` and `pkgs.writeShellApplication` not working correctly.


rvem (approved):
Thanks!

PR #295, opened 2024-09-25
doc: add option doc for interactiveSudo

AsterisMono:
Password-based sudo is implemented in #257 but not documented in the README.

See also: #78


rvem (approved):
Thanks!

Issue #260, opened 2024-02-16
Request: Show failed nix command


tecosaur:
For some reason, my nix daemon has started having "Connection refused issues" during attempted deployments

```
🚀 ℹ️ [deploy] [INFO] Building profile `system` for node `golgi`
🚀 ℹ️ [deploy] [INFO] Copying profile `system` to node `golgi`
error: cannot connect to socket at '/nix/var/nix/daemon-socket/socket': Connection refused
error: unexpected end-of-file
🚀 ❌ [deploy] [ERROR] Failed to push profile: Nix copy command resulted in a bad exit code: Some(1)
```

I'm not sure where this issue is coming from, but one thing that might make it easier to track down is if deploy-rs were to print the `nix` command that failed when it produces a bad exit code.

heitor-lassarote:
Resolved by #310.

PR #310, opened 2025-01-10
[#260] Show failed nix command

heitor-lassarote:
This PR displays the command that failed when an error is produced after a command exits with a non-zero code. The output is also cleared a bit. Example:

```
🚀 ℹ️ [deploy] [INFO] Running checks for flake in examples/simple/flake.nix
🚀 ❌ [deploy] [ERROR] Failed to check deployment: Nix checking command resulted in a bad exit code: Some(1). The failed command is provided below:
Command { std: "nix" "flake" "check" "examples/simple/flake.nix", kill_on_drop: false }
The stderr output is provided below:
error: in flake URL 'examples/simple/flake.nix', 'flake.nix' is not a commit hash
```

### DIFF
--- a/src/bin/activate.rs
+++ b/src/bin/activate.rs
@@ -125,53 +125,57 @@ struct RevokeOpts {
 pub enum DeactivateError {
     #[error("Failed to execute the rollback command: {0}")]
     Rollback(std::io::Error),
-    #[error("The rollback resulted in a bad exit code: {0:?}")]
-    RollbackExit(Option<i32>),
+    #[error("The rollback resulted in a bad exit code: {0:?}. The failed command is provided below:\n{1}")]
+    RollbackExit(Option<i32>, String),
     #[error("Failed to run command for listing generations: {0}")]
     ListGen(std::io::Error),
-    #[error("Command for listing generations resulted in a bad exit code: {0:?}")]
-    ListGenExit(Option<i32>),
+    #[error("Command for listing generations resulted in a bad exit code: {0:?}. The failed command is provided below:\n{1}")]
+    ListGenExit(Option<i32>, String),
     #[error("Error converting generation list output to utf8: {0}")]
     DecodeListGenUtf8(std::string::FromUtf8Error),
     #[error("Failed to run command for deleting generation: {0}")]
     DeleteGen(std::io::Error),
-    #[error("Command for deleting generations resulted in a bad exit code: {0:?}")]
-    DeleteGenExit(Option<i32>),
+    #[error("Command for deleting generations resulted in a bad exit code: {0:?}. The failed command is provided below:\n{1}")]
+    DeleteGenExit(Option<i32>, String),
     #[error("Failed to run command for re-activating the last generation: {0}")]
     Reactivate(std::io::Error),
-    #[error("Command for re-activating the last generation resulted in a bad exit code: {0:?}")]
-    ReactivateExit(Option<i32>),
+    #[error("Command for re-activating the last generation resulted in a bad exit code: {0:?}. The failed command is provided below:\n{1}")]
+    ReactivateExit(Option<i32>, String),
 }
 
 pub async fn deactivate(profile_path: &str) -> Result<(), DeactivateError> {
     warn!("De-activating due to error");
 
-    let nix_env_rollback_exit_status = Command::new("nix-env")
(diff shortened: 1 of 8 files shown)


src/command.rs
```diff
@@ -0,0 +1,131 @@
...
+}
+
+impl Command {
+    pub fn new<S: AsRef<OsStr>>(program: S) -> Command {
+        Command {
+            command: TokioCommand::new(program),
+        }
+    }
+
+    pub fn arg<S: AsRef<OsStr>>(&mut self, arg: S) -> &mut Command {
```
rvem, line 57: What is the motivation behind the reimplementation of `arg` and other methods defined for `TokioCommand` rather than creating a single `new` constructor that accepts already constructed `TokioCommand` and `run` to run the command handling the result?

src/bin/activate.rs
```diff
@@ -362,17 +376,31 @@ pub async fn wait(temp_path: PathBuf, closure: String, activation_timeout: Optio
...
+pub enum RunActivateError {}
+
+impl command::HasCommandError for RunActivateError {
+    fn title() -> String {
+        "Nix activation script".to_string()
+    }
+}
+
 #[derive(Error, Debug)]
 pub enum ActivateError {
```
rvem, line 398: I like the fact that this error type became simpler

src/deploy.rs
```diff
@@ -409,23 +437,35 @@ pub async fn deploy_profile(
...
+                .map_err(|err| {
+                    DeployProfileError::SSHActivate(command::CommandError::OtherError(
+                        SSHActivateError::ActivatePipeError(err)
+                    ))
+                })?;
         }
 
         let ssh_activate_exit_status = ssh_activate_child
-            .wait()
+            .wait_with_output()
```
rvem, line 458: Is there a reason why the `run` method of `command::Command` is not used here instead of `wait_with_output`?

heitor-lassarote: Because `ssh_activate_child` is a `tokio::process::Child` and not a `command::Command` or `tokio::process::Command`. Although there are patterns, the way the codebase handles children is not repetitive enough to easily abstract them into a `run` command for a new `command::Child`, but I can put some more thought into this if you'd like.

src/command.rs
```diff
@@ -0,0 +1,72 @@
...
+                write!(f, "Failed to run {} command: {}", T::title(), err,)
+            }
+            CommandError::Exit(output, cmd) => {
+                let stderr = match String::from_utf8(output.stderr.clone()) {
+                    Ok(stderr) => stderr,
+                    Err(_err) => format!("{:?}", output.stderr),
+                };
+                write!(
+                    f,
+                    "{} command resulted in a bad exit code: {:?}. The failed command is provided below:\n{}\nThe stderr output is provided below:\n{}",
```
Sereja313, line 30: When nix source code is embedded into a command, it also embeds the indentation:
```
🚀 ❌ [deploy] [ERROR] Failed to evaluate deployment data: Nix eval command resulted in a bad exit code: Some(1). The failed command is provided below:
Command { std: "nix" "eval" "--json" ".#deploy" "--apply" "\n                      deploy:\n                      (deploy // {\n                        nodes = {\n                          inherit (deploy.nodes) \"skat\";\n                        };\n                      })\n                    ", kill_on_drop: false }
```
Perhaps we could use a heuristic algorithm that replaces a \n character followed by space characters with a single space. So it would like smth like:
```
Command { std: "nix" "eval" "--json" ".#deploy" "--apply" " deploy: (deploy // { nodes = { inherit (deploy.nodes) \"skat\"; }; }) ", kill_on_drop: false }
```

OTOH, this is debugging information, so consistency with the original may be more important than brevity, but since nix is ​​not sensitive to indentation this heuristic shouldn't cause any problems.

What do you think?

heitor-lassarote: By the way, another alternative is to unescape characters, so `\n` for example turns into an actual newline and indentations keep making sense. What would you prefer?

heitor-lassarote: Pushed a new commit improving the formatting.

heitor-lassarote: Oops, committed a test example for accident, I deleted it in a fixup commit.

Sereja313 (approved):

PR #314, opened 2025-02-01
refactor: remove flake-utils

isabelroses:
its only used for generating system dependant outputs. it would have made more sense to use if systems were provided as a systems were provided as an input.

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -35,41 +35,7 @@
     "root": {
       "inputs": {
         "flake-compat": "flake-compat",
-        "nixpkgs": "nixpkgs",
-        "utils": "utils"
-      }
-    },
-    "systems": {
-      "locked": {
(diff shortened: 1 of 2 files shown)


jalil-salame (approved):

trueNAHO (changes requested):
> it would have made more sense to use if systems were provided as a systems were provided as an input.

The currently pending commit https://github.com/serokell/deploy-rs/pull/345/commits/428b3a54cabd0fd93fa487979e225c43de679785 ("Add systems input to leverage extensible flake systems pattern") from https://github.com/serokell/deploy-rs/pull/345 does exactly that. Thus, superseding this PR.

Interfacing the systems input is required to resolve https://github.com/serokell/deploy-rs/pull/209.

Issue #307, opened 2024-12-25
Unrecognized node or token encountered


ReneHollander:
Hello!

I am very stumed about the following error when I try to deploy a specific node:

```
[rene@renespc:~/repositories/infra]$ nix run github:serokell/deploy-rs -- -d .#chwei01.serv0r.xyz
🚀 ❌ [deploy] [ERROR] Error parsing flake: Unrecognized node or token encountered

[rene@renespc:~/repositories]$ nix run github:serokell/deploy-rs -- -d infra#chwei01.serv0r.xyz
🚀 ❌ [deploy] [ERROR] Error parsing flake: Unrecognized node or token encountered
```

When I just deploy without trying to specify a target it works perfectly fine:
```
$ nix run github:serokell/deploy-rs
🚀 ℹ️ [deploy] [INFO] Running checks for flake in .
warning: Git tree '/home/rene/repositories/infra' is dirty
warning: unknown flake output 'homeManagerModules'
warning: unknown flake output 'deploy'
warning: unknown flake output 'terraformConfiguration'
warning: unknown flake output 'dockerContainers'
warning: The check omitted these incompatible systems: aarch64-darwin, aarch64-linux, x86_64-darwin
Use '--all-systems' to check all.
🚀 ℹ️ [deploy] [INFO] Evaluating flake in .
warning: Git tree '/home/rene/repositories/infra' is dirty
🚀 ℹ️ [deploy] [INFO] The following profiles are going to be deployed:
["chwei01.serv0r.xyz".system]
user = "root"
ssh_user = "nixbuilder"
path = "/nix/store/044zwkf1qi1bq6cifci5yy7k810kfcn4-activatable-nixos-system-chwei01-25.05.20241219.d70bd19"
hostname = "chwei01.serv0r.xyz"
ssh_opts = []
...
🚀 ℹ️ [deploy] [INFO] Building profile `system` for node `chwei01.serv0r.xyz`
...
🚀 ℹ️ [deploy] [INFO] Copying profile `system` to node `chwei01.serv0r.xyz`
...
```

I can't find any reference to the above error anywhere. Any hints as to what to check?

ReneHollander:
As per usual I figured it out a few minutes later: deploy-rs does not like the node name to contain `.`.

Updated the issue title accordingly.

jacks-sam1010:
Hello, we tried to solve the issue.

This is what we did:

[...]

PR #317, opened 2025-03-27
update flake and rust dependencies

Chaostheorie:
This is a naive update with minor adjustments for new APIs. Most changes are clap related because their macro naming has changed.


Chaostheorie:
The only skipped dependency is `rnix`, it had a few breaking API changes that likely require a bigger refactor. I can try to tackle this in the new few days.

Chaostheorie:
This also now contains a minor fix for #307, if required, this can be split into its own PR

Chaostheorie:
Tested the PR locally, works fine. No regression appears to be present.

Chaostheorie:
After an hour on a builder, nixos vm test failed, seems like activation CLI needs some more work:

```
vm-test-run-options-overriding> client # thread 'tokio-runtime-worker' panicked at /build/source/src/deploy.rs:493:37:
vm-test-run-options-overriding> client # called `Result::unwrap()` on an `Err` value: ()
```

weriomat (commented):
Hey, I have debugged the issue with the tests.
Since you bumped the nixpkgs version, [this](https://github.com/NixOS/nixpkgs/commit/4cfbbb3c1ded303eccb09d53319dfb1cd862bfdc) commit removes the `switch-to-configuration` script by default.
We should enable this option in the common config because the test VMs need to have the new rust script in path (since they cannot fetch it while being run)

weriomat (changes requested):
The `overlay` outputs seems to be deprecated as well in favor of `overlays.default`

weriomat:
@rvem the build failures for on the `automatic-updates` branch are also fixed here :) Thus a review from you would be appreciated.

flake.nix
```diff
@@ -16,9 +16,8 @@
   };
 
   outputs = { self, nixpkgs, utils, ... }@inputs:
-  rec {
-    overlay = final: prev: let
```
rvem, line 20: Can we preserve the `overlay` output for the sake of backward compatibility?
Just leave it as `overlay = overlays.default`

src/lib.rs
```diff
@@ -231,6 +231,7 @@ pub fn parse_flake(flake: &str) -> Result<DeployFlake, ParseFlakeError> {
 
                     Some(c.into_token().unwrap().text().to_string())
                 }
+                (NODE_SELECT, _) => Some(entry.into_node().unwrap().text().to_string()),
```
rvem, line 234: Could you add a test case with the node that has `.` in the name?

rvem: Hmm, I tried, to add `local.server` to the node attrset and deploy it with `deploy -s .#local.server`, and got
```
vm-test-run-node-name-with-dot> client #        error: attribute 'profiles' missing
vm-test-run-node-name-with-dot> client #        at «string»:7:40:
vm-test-run-node-name-with-dot> client #             6|                             profiles = {
vm-test-run-node-name-with-dot> client #             7|                               inherit (deploy.nodes."local".profiles) "server";
vm-test-run-node-name-with-dot> client #              |                                        ^
vm-test-run-node-name-with-dot> client #             8|                             };
```

rvem (changes requested):
Changes except for https://github.com/serokell/deploy-rs/pull/317/commits/0c2b81d25c1d35d29b44094b41496053f8eef1aa look good to me.
Dotted node names seem to be trickier, thus I propose to deal with it in a separate PR

src/lib.rs
weriomat: When I took a look at the issue at hand, deploy-rs parses the flake uri as  `<path>#<node>.<profile>` and therefore it is not accepted (take a look at the [tests](https://github.com/serokell/deploy-rs/blob/master/src/lib.rs#L299) for this function). In this case we try to deploy the node `local` with the profile `server`. The README contains examples for this issue. If you try to deploy `local.server` as the node you have to use `.#"local.server"` as specified in the README. Therefore, it is not accepted by design.

weriomat: In my opinion 0c2b81d should be reverted

PR #320, opened 2025-05-04
`--file` argument analogous to `nix build` & friends

stuebinm:
This adds a `--file` option, which allows using `deploy` in repositories which are not nix flakes. `deploy-rs` already has support for falling back to a non-flake version of its logic in case the `nix` available in `$PATH` doesn't have the flake experimental feature flag enabled. However, that's an "if and only if"-situation: if the nix *is* flakes-enabled, there's no way to use the non-flake version. This works well for setups which use nix flakes, but provides a flake-compat `default.nix` fallback, but prevents use of `deploy-rs` in setups which do not use flakes at all but rely on another source-managing tool such as `niv` or `npins`.

`--file` takes as argument a path, and `--targets` and other options continue to work as expected; remote builds also work assuming a new enough nix on the target host. I've also added tests for local & remote builds (and fixed the name of the `non-flake-build` test; looks like i forgot that when I introduced it in https://github.com/serokell/deploy-rs/pull/272)

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -30,6 +30,9 @@ pub struct Opts {
     /// A list of flakes to deploy alternatively
     #[clap(long, group = "deploy")]
     targets: Option<Vec<String>>,
+    /// Treat targets as files instead of flakes
+    #[clap(short, long)]
+    file: Option<String>,
     /// Check signatures when using `nix copy`
     #[clap(short, long)]
     checksigs: bool,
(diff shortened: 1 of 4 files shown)


src/cli.rs
```diff
@@ -672,15 +675,28 @@ pub async fn run(args: Option<&ArgMatches>) -> Result<(), RunError> {
         error!("Cannot use both --dry-activate & --boot!");
     }
 
+    // if opts.file.is_some() && (opts.targets.is_some() || opts.target.is_some()) {
```
rvem, line 678: Is this a leftover?

src/push.rs
```diff
@@ -169,7 +167,9 @@ pub async fn build_profile_remotely(data: &PushProfileData<'_>, derivation_name:
 
 
     // copy the derivation to remote host so it can be built there
-    let copy_command_status = Command::new("nix").arg("copy")
+    let copy_command_status = Command::new("nix")
+        .arg("--experimental-features").arg("nix-command")
```
rvem, line 171: I wonder if it makes sense to add `--experimental-features=nix-command` only if `!data.supports_flakes`, what do you think?

rvem (approved):
LGTM, thanks for your contribution

PR #327, opened 2025-06-06
Fix multiple targets failing to be parsed after a major clap update (fixes #325)

### LINKED ISSUE #325: a6b0cb2c5e1c1efea2d48fdf7bb5b36b6675e40e broke `--targets`
Single argument to `--targets` is still fine, but multiple targets are not.  Now when you say `deploy --targets .#{host-a,host-b}.system` you get:
```
error: the argument '--targets <TARGETS>' cannot be used with '[TARGET]'

Usage: deploy --targets <TARGETS> [TARGET] <EXTRA_BUILD_ARGS>...
```

The commit in question "just" updates deps, perhaps the Clap major-version bump is the problem?

VolodiaPG:
This fixes #325 by adding a missing argument to the decorator of the targets cli arg.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -28,7 +28,7 @@ pub struct Opts {
     target: Option<String>,
 
     /// A list of flakes to deploy alternatively
-    #[arg(long, group = "deploy")]
+    #[arg(long, group = "deploy", num_args = 1..)]
     targets: Option<Vec<String>>,
     /// Treat targets as files instead of flakes
     #[clap(short, long)]


sedlund:
build args are still broken with this

> nix run "github:VolodiaPG/deploy-rs?rev=2edaf49" -- --targets .#wsl.system -- -Lv

```
error: the argument '--targets <TARGETS>...' cannot be used with '[TARGET]'

Usage: deploy --targets <TARGETS>... [TARGET] [EXTRA_BUILD_ARGS]...
```

thanks for looking at it

heitor-lassarote:
Thanks for finding this and for the `num_args = 1..` fix. That was the right diagnosis.

#341 built on it and also fixed the extra-build-args case reported here, and it's merged on `master`. #325 is closed because of it.

Closing this one since master already has the fix.

PR #331, opened 2025-06-14
chore: respect TMPDIR for local builds

### LINKED ISSUE #283: `$TMPDIR` env var not entirely respected in local builds
I effectively cannot build any configuration when the build machine has `/tmp` mounted on a partition (or `tmpfs` in my case) when there isn't enough space for the build (which is a rather common occurrence lately). Same as with [remote builds](https://github.com/serokell/deploy-rs/issues/282) when the _target_ machine is limited similarly.

The workaround for `nixos-rebuild` is to set `$TMPDIR` to somewhere with enough capacity for the build but this is not _entirely_ respected by `deploy` as a part of the build does happen in whatever `TMPDIR` is set to but the "big" build still happens in `/tmp` either locally or on the remote machine, depending on `remoteBuild`.

Example error for a local build, `TMPDIR` set to `/tmp2`:
```
🚀 ℹ️ [deploy] [INFO] Building profile `system` for node `muffy`
error:
       … writing file '/tmp/nix-build-etc-alsa-conf.d-49-pipewire-modules.conf.drv-0/.attr-0l2nkwhif96f51f4amnlf414lhl4rv9vh8iffyp431v6s28gsr90'

       error: writing to file: No space left on device
🚀 ❌ [deploy] [ERROR] Failed to push profile: Nix build command resulted in a bad exit code: Some(1)
```

For local builds, `TMPDIR` should imho be respected in all build phases. For remote builds see #282.

The only workaround that _works for me_ (tm) is to remount `/tmp` to some place with enough capacity for the duration of the build, which is ... not optimal.

weriomat:
This fixes #283.

### DIFF
--- a/src/push.rs
+++ b/src/push.rs
@@ -132,6 +132,13 @@ pub async fn build_profile_locally(
         build_command.arg(derivation_name)
     };
 
+    match std::env::var("TMPDIR") {
+        Ok(build_dir) => {
+            info!("Detected TMPDIR is set for build to {build_dir}");
+            build_command.env("TMPDIR", build_dir);
+        }
+        Err(_) => {}
+    }
     match (data.keep_result, data.supports_flakes) {
         (true, _) => {
             let result_path = data


heitor-lassarote:
Thank you for this fix. I tried reproducing this bug locally, and I think this fix matters a bit less now.

Nix changed this on purpose: commit 88b7db1b ("Don't default build-dir to temp-dir, store setting", March 2025) moved the default build directory from `$TMPDIR` to a subdirectory of `nixStateDir`, as a hardening measure against sandbox escapes. So before that commit, `TMPDIR` was respected by default, which is likely why this fix worked when you wrote it.

On current Nix, for a multi-user (daemon) install, the daemon uses `<stateDir>/builds` (like `/nix/var/nix/builds`) and `TMPDIR` has no effect anymore. I confirmed this on both current Nix and the 2.24.12 this repo's nixpkgs pins. `--option build-dir <path>` still works there.

For a single-user (non-daemon) install, `TMPDIR` is still respected, but `Command` already inherits the parent's environment by default, so that case should work regardless of this change.

Could you share what setup and Nix version you tested against? Want to make sure I'm not missing something before deciding how to move forward here.

weriomat:
I have not checked in with this PR in a long time. With this PR I wanted to help out the OP in #283 but never ran into this problem myself.

As I only use multi user installs I cant test the single user one. Thus I am closing this PR.

Thank you for looking at this PR :)

PR #333, opened 2025-07-10
fix targets and build args from clap v4 update

sedlund:
Disclaimer: this was 'vibe' 🤗 patched with LLM, I know very little of rust.

this fixes usage after the clap v4 update.

it removes `--targets` and allows `--target` to be repeated as is how I understand clap v4 wants to work.  (having both was confusing and redundant anyway)

to easily specify multiple targets utilize shell brace expansion:

`--target=.#{host1.profile1,host2.profile3}` etc...

Everything seems to work for me so far.  YMMV

#325

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,7 +28,7 @@ Any "extra" arguments will be passed into the Nix calls, so for instance to depl
 You can try out this tool easily with `nix run`:
 - `nix run github:serokell/deploy-rs your-flake`
 
-If you want to deploy multiple flakes or a subset of profiles with one invocation, instead of calling `deploy <flake>` you can issue `deploy --targets <flake> [<flake> ...]` where `<flake>` is supposed to take the same format as discussed before.
+If you want to deploy multiple flakes or a subset of profiles with one invocation, instead of calling `deploy <flake>` you can add `--target` multiple times.  Use shell brace expansion to make this easier: `deploy --target={<flake>,<flake>,...}` where `<flake>` is supposed to take the same format as discussed before. Ex: `--target=.#{host1,host2.system,host3.myuser}`.
 
 Running in this mode, if any of the deploys fails, the deploy will be aborted and all successful deploys rolled back. `--rollback-succeeded false` can be used to override this behavior, otherwise the `auto-rollback` argument takes precedent.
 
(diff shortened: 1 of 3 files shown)


weriomat:
You should probably update the `interface.json` accordingly

sedlund:
Thanks.  I updated the test to use `--target`.
 I didn't see anything specific in `interface.json` that needs to change. 🧐

weriomat:
Sorry you are right, there is nothing to be changed there.
Regardless in my opinion this introduces a regression.
I can no longer specify a target via `deploy -d -s .#test` but I need to `deploy -d -s --target .#test`. 
I can no longer specify multiple targets, but I need to invoke `--target` multiple times. 
Fixing your problem can be done with only removing the `group = "deploy"` from `target` and `targets`.
Maybe we should use [custom Validation](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#custom-validation) or no validation at all?

sedlund:
I appreciate your feedback.

>  I can no longer specify a target via `deploy -d -s .#test`

This was unintentional.

Although, having 3 ways to specify hosts to deploy is redundant in the code and mental space.  It makes the tool less approachable.

I suggest having one way to pass targets in the future, however that may be (_while supporting brace expansion_).

> I can no longer specify multiple targets, but I need to invoke `--target` multiple times

[Shell brace expansion](https://www.linuxjournal.com/content/bash-brace-expansion) makes this better than not using them in most cases.  This was possible before.

For example, with the original `--targets` option:

`
--targets myFlake#hostAlpha.alice myFlake#hostAlpha.bob myFlake#hostBravo.alice myFlake#hostBravo.bob myFlake#hostCharlie.alice myFlake#hostCharlie.bob
`

becomes:

`--target=myFlake#host{Alpha,Bravo,Charlie}.{alice,bob}`

> Fixing your problem can be done with only removing the `group = "deploy"` from `target` and `targets`. 
Maybe we should use [custom Validation](https://docs.rs/clap/latest/clap/_derive/_tutorial/index.html#custom-validation) or no validation at all?

For clarity, the issues with deploy-rs HEAD as well as in stable `nixos-25.05`:
1. Cannot specify multiple targets.
2. Cannot pass nix build args.

This PR allows a ref to patch with to resolve these (my goal).

I don't have time to continue without concrete suggestions for improvement.  If something has been merged to resolve, I will close this.

Thanks again.

PR #341, opened 2025-10-01
Fix targets and extra build args

### LINKED ISSUE #325: a6b0cb2c5e1c1efea2d48fdf7bb5b36b6675e40e broke `--targets`

sjagoe:
This fixes #325 by building upon the fix (#327) by @VolodiaPG for build targets, adding a fix for extra build args.

With this, deploying to multiple hosts with build args works again, e.g.

```
../deploy-rs/target/debug/deploy -d --targets .#host1 .#host2 .#host3 --keep-result --skip-checks -- --max-jobs 0
```

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -40,6 +40,7 @@ pub struct Opts {
     #[arg(short, long)]
     interactive: bool,
     /// Extra arguments to be passed to nix build
+    #[arg(last = true)]
     extra_build_args: Vec<String>,
 
     /// Print debug logs to output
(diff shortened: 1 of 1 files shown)


Sntx626 (approved):
Tested, fixes the problem, looks good.
LGTM

neunenak:
I ended up pushing a PR to fix this https://github.com/serokell/deploy-rs/pull/351 without checking to see if that had been done already. @Sntx626 what needs to happen for this fix to be mergeable?

PR #346, opened 2025-10-19
fix compatibility with Nix 2.32+ show-derivation output

### LINKED ISSUE #340: Cannot deploy with Nix 2.32(pre)
I have installed nix from main branch of nix, nix 2.32-pre, and I couldn't deploy because of a strange error (no such file pointing to cwd/<hash>-activatable-nixos-system-...-<version>.drv), I narrowed it down to the fact that `nix show-derivation` no longer shows full path with newer nix (I am not sure when the change was introduced, reverted back to 2.28 and it's fine)

Instead of something like
```
> nix show-derivation /nix/store/m1sczr2hq5z1xkg2hzhlpjq0c42isdii-activatable-nixos-system-cerebrum-25.11.20250916.8d4ddb1
{
  "/nix/store/22s18pb08i691lnal0pfm7yyf40pqsdv-activatable-nixos-system-cerebrum-25.11.20250916.8d4ddb1.drv": {
```

I am getting
```
> nix derivation show /nix/store/m1sczr2hq5z1xkg2hzhlpjq0c42isdii-activatable-nixos-system-cerebrum-25.11.20250916.8d4ddb1
{
  "22s18pb08i691lnal0pfm7yyf40pqsdv-activatable-nixos-system-cerebrum-25.11.20250916.8d4ddb1.drv": {
```

This path is obtained in deploy-rs code and then put to `nix path-info`. 1. it's interpreted as local path, searching under cwd, 2. even if in nix store, there is an error that the path is not a flake.

Here is the problematic code https://github.com/serokell/deploy-rs/blob/125ae9e3ecf62fb2c0fd4f2d894eb971f1ecaed2/src/push.rs#L223-L261

XYenon:
Fixes #340

### DIFF
--- a/src/push.rs
+++ b/src/push.rs
@@ -240,17 +240,25 @@ pub async fn build_profile(data: PushProfileData<'_>) -> Result<(), PushProfileE
     )
     .map_err(PushProfileError::ShowDerivationParse)?;
 
-    let &deriver = derivation_info
+    let deriver_key = derivation_info
         .keys()
         .next()
         .ok_or(PushProfileError::ShowDerivationEmpty)?;
 
-    let new_deriver = &if data.supports_flakes || data.deploy_data.merged_settings.remote_build.unwrap_or(false) {
+    // Nix 2.32+ returns relative paths (without /nix/store/ prefix) in show-derivation output
+    // Normalize to always use full store paths
+    let deriver = if deriver_key.starts_with("/nix/store/") {
+        deriver_key.to_string()
+    } else {
+        format!("/nix/store/{}", deriver_key)
+    };
+
+    let new_deriver = if data.supports_flakes || data.deploy_data.merged_settings.remote_build.unwrap_or(false) {
         // Since nix 2.15.0 'nix build <path>.drv' will build only the .drv file itself, not the
         // derivation outputs, '^out' is used to refer to outputs explicitly
-        deriver.to_owned().to_string() + "^out"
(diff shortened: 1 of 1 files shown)


notgne2:
Well I can't merge anymore apparently, but _please_, @rvem ? @PhilTaken ? anybody

rvem (approved):

src/push.rs
```diff
@@ -240,17 +240,25 @@ pub async fn build_profile(data: PushProfileData<'_>) -> Result<(), PushProfileE
...
         .next()
         .ok_or(PushProfileError::ShowDerivationEmpty)?;
 
-    let new_deriver = &if data.supports_flakes || data.deploy_data.merged_settings.remote_build.unwrap_or(false) {
+    // Nix 2.32+ returns relative paths (without /nix/store/ prefix) in show-derivation output
+    // Normalize to always use full store paths
+    let deriver = if deriver_key.starts_with("/nix/store/") {
+        deriver_key.to_string()
+    } else {
+        format!("/nix/store/{}", deriver_key)
```
Rutherther: This doesn't allow locations for nix store other than /nix/store. One solution would be to look at $NIX_STORE_DIR env var for location of the store and prepend that, only defaulting to /nix/store. Another would be to utilize the output of `nix eval --expr builtins.storeDir`.

Rutherther (commented):
Changes from this PR will break usage for people using nix store at different location than the default one.

PR #347, opened 2025-10-22
use shlex to parse/ escape stings for POSIX shells

### LINKED ISSUE #130: wrong arguments given to ssh when some ssh argument specified in ssh-opts contains space
Running ` deploy --ssh-opts='-o ProxyCommand="wstunnel --upgradePathPrefix=simple-test -L stdio:%h:%p wss://wstunnel.example.com"'  '.#test'` passes `["-o", "ProxyCommand=\"wstunnel", "--upgradePathPrefix=simple-test", "-L", "stdio:%h:%p", "wss://wstunnel.example.com\""]` to ssh.

The current implementation just splits spaces.
https://github.com/serokell/deploy-rs/blob/e5546f9c2503c26d175f08a81fc0a0f330be4cbe/src/lib.rs#L418

Instead, we should do something like [shlex.split](https://docs.python.org/3/library/shlex.html#shlex.split). I expect the arguments given to ssh is `ssh_opts = ['-o', 'ProxyCommand=wstunnel --upgradePathPrefix=simple-test -L stdio:%h:%p wss://wstunnel.example.com']` instead of `ssh_opts = ["-o", "ProxyCommand=\"wstunnel", "--upgradePathPrefix=simple-test", "-L", "stdio:%h:%p", "wss://wstunnel.example.com\""]`.

This may be usefule. https://users.rust-lang.org/t/crate-for-splitting-a-string-like-bash/40062

### LINKED ISSUE #343: Working example of configuration with SSH and ProxyCommand
Hello,

Today I tried to use `deploy-rs`, my favorite deployment tool to provision an EC2 instance. To access this instance, I must authenticate through a proxy, therefore, my `.ssh/config` contains something like:

```
Host foobar
  User root
  Hostname 127.0.0.1
  IdentitiesOnly yes
  IdentityFile ~/.ssh/id_foobar
  ProxyCommand ncat --proxy-auth proxyUser:proxyPassword --proxy proxyHost:proxyPort %h %p 
```

This allow me to connect to the box just by doing `ssh foobar`. This works pretty well.

However, when it comes to `deploy-rs`, things are a bit more complex.

Here's my deploy node configuration:

```nix
{
  deploy.nodes.foobar = {
    hostname = "127.0.0.1";
    fastConnection = false;
    profiles.system = {
      remoteBuild = true;
      sshUser = "root";
      sshOpts = [
        "-o"
        "ProxyCommand=ncat --proxy-auth proxyUser:proxyPassword --proxy proxyHost:proxyPort %h %p"
        "-i"
        "/home/pol/.ssh/id_foobar"
      ];
      path = inputs.deploy-rs.lib.x86_64-linux.activate.nixos inputs.self.nixosConfigurations.foobar;
    };
  };
}
```

When I execute this, I get:

```
❯ deploy -s .#foobar
🚀 ℹ [deploy] [INFO] Evaluating flake in .
🚀 ℹ [deploy] [INFO] The following profiles are going to be deployed:
[foobar.system]
user = "root"
ssh_user = "root"
path = "/nix/store/sbkwmlwmiql9v9nksfi8hckdic3ks7r0-activatable-nixos-system-foobar-25.11.20251012.cf3f5c4"
hostname = "127.0.0.1"
ssh_opts = ["-o", 'ProxyCommand="ncat --proxy-auth foo:bar --proxy redacted:8012 %h %p"', "-i", "/home/pol/.ssh/id_foobar"]

🚀 ℹ [deploy] [INFO] Building profile `system` for node `foobar` on remote host
🚀 ℹ [deploy] [INFO] Activating profile `system` for node `foobar`
🚀 ℹ [deploy] [INFO] Creating activation waiter
/nix/store/cl2gkgnh26mmpka81pc2g5bzjfrili92-bash-5.3p3/bin/bash: line 1: exec: ncat --proxy-auth foo:bar --proxy redacted:8012 ip 22: not found
Connection closed by UNKNOWN port 65535
🚀 ❌ [deploy] [ERROR] Activating over SSH resulted in a bad exit code: Some(255)
🚀 ℹ [deploy] [INFO] Revoking previous deploys
🚀 ❌ [deploy] [ERROR] Deployment to node foobar failed, rolled back to previous generation
/nix/store/cl2gkgnh26mmpka81pc2g5bzjfrili92-bash-5.3p3/bin/bash: line 1: exec: ncat --proxy-auth foo:bar --proxy redacted:8012 IP 22: not found
Connection closed by UNKNOWN port 65535
```

I tried many combinations and I couldn't find anything working yet. I have the feeling that this is related to https://github.com/serokell/deploy-rs/issues/130

Do you have a clue to share?

Using `nixos-rebuild switch --flake .#foobar --target-host foobar --build-host foobar` works without any trouble. I guess this is because it's using my local SSH configuration.

weriomat:
Use the [shlex](https://crates.io/crates/shlex) crate to split/ parse the `ssh_opts` instead of just splitting by `' '`.
If shlex fails to parse/ join, we fall back to the old way.

Fixes #130. Might fix #343.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,7 @@ dependencies = [
  "rpassword",
  "serde",
  "serde_json",
+ "shlex",
  "signal-hook",
  "thiserror 2.0.12",
  "tokio",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -32,6 +32,7 @@ rpassword = "7.3.1"
 unescape = "0.1.0"
 indicatif = { version = "0.17.8", features = ["tokio", "futures"] }
 tokio-stream = { version = "0.1.16", features = ["io-util"] }
+shlex = "1.3.0"
 
 
 [lib]
--- a/src/lib.rs
+++ b/src/lib.rs
@@ -20,6 +20,13 @@ pub fn make_lock_path(temp_path: &Path, closure: &str) -> PathBuf {
     temp_path.join(format!("deploy-rs-canary-{}", lock_hash))
 }
 
+/// Splits `--ssh-opts` like a POSIX shell, so a quoted option isn't broken
+/// apart on its own spaces. Falls back to a plain space split if the
+/// string isn't validly quoted.
+fn parse_ssh_opts(ssh_opts: &str) -> Vec<String> {
+    shlex::split(ssh_opts).unwrap_or(ssh_opts.split(' ').map(|x| x.to_owned()).collect())
+}
+
 const fn make_emoji(level: log::Level) -> &'static str {
     match level {
         log::Level::Error => "❌",
@@ -533,8 +540,7 @@ pub fn make_deploy_data(
         merged_settings.user = cmd_overrides.profile_user.clone();
     }
     if let Some(ref ssh_opts) = cmd_overrides.ssh_opts {
-        merged_settings.ssh_opts =
-            shlex::split(ssh_opts).unwrap_or(ssh_opts.split(' ').map(|x| x.to_owned()).collect());
+        merged_settings.ssh_opts = parse_ssh_opts(ssh_opts);
     }
     if let Some(fast_connection) = cmd_overrides.fast_connection {
         merged_settings.fast_connection = Some(fast_connection);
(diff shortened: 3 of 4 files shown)


heitor-lassarote:
Thank you for your work on this. I've rebased your branch (and resolved conflicts) and added some tests.

One note: I don't think this fixes #343. That failure happens during the direct SSH activation call, where each `ssh_opts` entry is already passed as its own argument, not through a joined string. The two spots this PR touches (the `--ssh-opts` CLI flag and NIX_SSHOPTS for remote builds) aren't on that code path. I'll leave #343 open separately.

CI is currently red. I didn't delve into it myself but I had Claude check. So please don't take its feedback at face value, just something that perhaps could help unblock this PR:

---

`join_ssh_opts` (used at both `NIX_SSHOPTS` call sites in push.rs) uses `shlex::try_join`, which quotes more aggressively than it needs to. It wraps something like `StrictHostKeyChecking=no` in single quotes even though there's no space to protect.

That's a problem because Nix's `NIX_SSHOPTS` parsing only became quote-aware in Nix 2.26. Before that, Nix split the string with `tokenizeString`, a plain whitespace split with no quote or escape handling at all (confirmed in Nix 2.24.12's own source, `SSHMaster::addCommonSSHOpts`). This repo's pinned nixpkgs ships Nix 2.24.12, so the quoted value comes through as one token, quotes included, and ssh chokes on it: `Bad configuration option: 'stricthostkeychecking`. That matches the failing jobs exactly.

It's not limited to ProxyCommand-style values either. Since shlex quotes almost anything containing `=`, it affects nearly every ssh option, which is why most of the VM tests fail, not just the one aimed at a specific edge case.

A quoting function that only wraps a value in quotes when it actually contains whitespace, instead of delegating to shlex's more conservative default, would fix the space-in-a-value case on Nix 2.26+ without changing anything for values that already worked, on any Nix version. Worth noting: even with that fix, a value with an embedded space still won't work on Nix below 2.26, since that Nix has no way to receive it correctly at all. That's a Nix-side limitation, not something fixable from here.

---

I don't have the bandwidth to implement that fix as part of this review, so flagging it here.

PR #357, opened 2026-01-02
Fix RunDeployError user-facing error message

neunenak:
The error macro used to display a human-readable string associated with a RunDeployError had what looks like a copy-paste error - it was using `{0}` twice to substitute in the string representation of the first argument of the error type; so it as printing that string twice redundantly and not printing the second argument at all.


weriomat (approved):

heitor-lassarote:
Thanks for catching this. This was fixed on master in commit b0eada8 (#260). Closing this as already resolved, but appreciate the report.

PR #359, opened 2026-01-04
fix compatibility with nix 2.33+ show-derivation output

szlend:
Broken by this change: https://nix.dev/manual/nix/2.33/release-notes/rl-2.33.html#derivation-json-format-changes

Fixes #355

### DIFF
--- a/src/push.rs
+++ b/src/push.rs
@@ -3,7 +3,6 @@
 // SPDX-License-Identifier: MPL-2.0
 
 use log::{debug, info, warn};
-use std::collections::HashMap;
 use std::path::Path;
 use std::process::Stdio;
 use thiserror::Error;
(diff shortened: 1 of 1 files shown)


src/push.rs
```diff
@@ -218,13 +219,10 @@ pub async fn build_profile(data: PushProfileData<'_>) -> Result<(), PushProfileE
     );
 
     // `nix-store --query --deriver` doesn't work on invalid paths, so we parse output of show-derivation :(
-    let mut show_derivation_command = Command::new("nix");
-
-    show_derivation_command
+    let show_derivation_output = Command::new("nix")
+        .arg("--experimental-features").arg("nix-command")
```
szlend, line 223: Not really a necessary change, but `show-derivation` is technically part of the `nix-command` experimental feature, and this is done in other places.

Also `nix show-derivation` has been deprecated by `nix derivation show` a while back, but not sure when we actually want to move to it. I left it alone for now as the old command still works.

ppenguin:
Thanks! I tested this and for me this worked and solved #355

topongo:
i'm also using this by default for some days and i didn't encounter any issues (and it saved my pants as i `pacman -Syu` too boldly)

szlend:
cc @rvem

georgeday-firefly:
working well for me

0rvar:
This PR fixed deploys for me as well. Good stuff!

Before:
```
error: getting status of '/nix/store/derivations': No such file or directory
```

After, deploy works fine

Sereja313 (approved):
Thanks

PR #361, opened 2026-02-01
flake: fix string escapes

antonmosich:
In nixlang "\\." is escaped to just ".". That is, it does nothing. For there to be a "\\." in the resulting string, two backslashes are needed. Lix on main now emits a warning in such cases. See the example nix repl output:

	nix-repl> :p "Cargo\.lock"
	Cargo.lock

	nix-repl> :p "Cargo\\.lock"
	Cargo\.lock

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,11 +26,11 @@
           version = "0.1.0";
 
           src = final.lib.sourceByRegex ./. [
-            "Cargo\.lock"
-            "Cargo\.toml"
+            "Cargo\\.lock"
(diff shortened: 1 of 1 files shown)


stuebinm:
Would be cool if this could be merged!

blackxored:
Lix 2.95 has been released so more users are seeing this. Just an FYI.

Luflosi:
@antonmosich could you please rebase this PR? I'd like to use it as my flake input.

PR #365, opened 2026-03-02
feat: add group-based deployment filtering

pippaye:
## Summary
  Add group-based filtering for deploy targets to allow partial deployments by logical groups.

## Changes
  - Add `groups` generic option (profile/node/deploy) with set-merge semantics.
  - Add CLI flag `--groups <GROUPS>...` and filter deployment selection.
  - Document usage and add an example under `examples/groups`.
  - Example added in `examples/groups`.

## Behavior
  - Effective groups are merged as a set from profile > node > deploy.
  - When `--groups` is provided, only profiles with any overlapping group are selected.
  - Profiles without groups are excluded when filtering is active.


Sereja313 (approved):
LGTM, thanks!

PR #367, opened 2026-03-16
remove unneeded nix string escapes

stuebinm:
These do not actually do anything, and are being deprecated in Lix; hence this change is a semantic no-op.

However, it might be that actually escaping the "." inside the regex was intended? (in that case, these should be `\\.`, although given that it seems to have never caused trouble, it's probably not really relevant)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -26,11 +26,11 @@
           version = "0.1.0";
 
           src = final.lib.sourceByRegex ./. [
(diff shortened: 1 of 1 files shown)


stuebinm:
Wait, someone already submitted this. Closing in favour of https://github.com/serokell/deploy-rs/pull/361

Issue #221, opened 2023-06-21
deploy-rs gives up on life after a single deployment fails


alxpettit:
I have like 8 machines to deploy to. One of them is offline. If deploy-rs tries to push to it first, it will fail, and then print `🚀 ❌ [deploy] [ERROR] Failed to push profile: Nix copy command resulted in a bad exit code: Some(1)`. No machines are updated.

Looking at the arg spec, I can't see any way to get it to continue pushing to the rest of the target machines.

jakehamilton:
I ran into this the other day. It would be great if there was a way to allow for skipping offline hosts explicitly while still failing on other deployment errors. Perhaps some opt-in flag like `--online-only` which skips the host if it's unreachable?

alxpettit:
My workflow currently involves a lot of experimental changes to very different machine profiles that can break on some sets of machines but not others. That being the case, I would prefer a mode that just keeps going until it's deployed everything it could, and reverted only the changes that would make machines inaccessible. :P

Ideally we'd have flags for `--skip-offline`, `--ignore-errors`, etc, which could be combined together into a "I don't care, just update everything that you can" mode. I believe there (probably) is a flag for the latter already, but it doesn't seem to actually work in my use case (e.g., deployment gives up when the nix profile tries to update vmware & a vmware service gets needlessly restarted, which reads as a deployment error and cancels the whole deployment).

Of course, that issue is outside the scope of this thread. I should probably check if an issue is already tracking it, and if not, make one.

notgne2:
~~this is intentionally modeled after my own behavior during deployments~~

It wasn't much of an issue initially, but there definitely needs to be a way to do this, I've had to comment things out in my home infra flake so frequently and it's very tiring.

I also do agree that if we add this, we probably need a `--idgaf` flag which would probably be something along the lines of `--auto-rollback false --magic-rollback false --skip-checks --skip-failed`, to give you the fastest and most reckless blasting of your deployments onto your machines possible.

dcurgz:
Hello, I needed this as well so I implemented `--skip-offline`, please try https://github.com/dcurgz/deploy-rs/tree/dcurgz/add-skip-offline if you still need it and I'll make a PR if it works OK.

cguentherTUChemnitz:
> Hello, I needed this as well so I implemented `--skip-offline`, please try https://github.com/dcurgz/deploy-rs/tree/dcurgz/add-skip-offline if you still need it and I'll make a PR if it works OK.

[...]

PR #370, opened 2026-04-15
add --skip-offline

jaspinegz:
Add a new option, --skip-offline, which partially addresses #221. 

If provided the CLI flag will attempt an SSH connection to each remote host in the deploy configuration. If an SSH connection cannot be established (5 second timeout) to a given host, the host is skipped.

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -558,6 +558,7 @@ async fn run_deploy(
             //https://stackoverflow.com/a/1405340
             let ssh_command_status = Command::new("ssh")
                 .arg("-q")
+                .arg("-o ConnectTimeout=5")
                 .arg(format!("ssh://{}@{}", deploy_defs.ssh_user, deploy_data.node.node_settings.hostname))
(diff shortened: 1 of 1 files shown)


src/cli.rs
```diff
@@ -549,6 +554,26 @@ async fn run_deploy(
...
+            //https://stackoverflow.com/a/1405340
+            let ssh_command_status = Command::new("ssh")
+                .arg("-q")
+                .arg("-o ConnectTimeout=5")
+                .arg(format!("ssh://{}@{}", deploy_defs.ssh_user, deploy_data.node.node_settings.hostname))
+                .arg("exit")
+                .stdout(Stdio::null())
+                .status()
+                .await
+                .unwrap();
```
Sntx626, line 567: Do we want to consider keeping the ssh connection alive and re-using it for the deployment? In case the connections is flakey and _some_ connection attempts fail. Since it could still result in hosts not being reachable during deployment...

PR #373, opened 2026-04-28
fix: cancel `wait` process on activation failure

### LINKED ISSUE #372: deploy hangs for `activation_timeout` when activation fails (`magic_rollback` enabled)
When `magic_rollback` is enabled and an activation script fails on the target, `deploy` hangs for the full `activation_timeout` (default 240s) before reporting the failure.

### Reproducing

Deploy any profile whose activation script exits non-zero. Example custom profile:

```nix
profiles.system.path = let
  failingActivate = pkgs.writeShellScriptBin "activate" "exit 1";
in deploy-rs.lib.${system}.activate.custom failingActivate "$PROFILE/bin/activate";
```

Run deploy against it. Observed sequence:

```
👀 ℹ️ [wait] [INFO] Waiting for confirmation event...
⭐ ⚠️ [activate] [WARN] De-activating due to error
🚀 ❌ [deploy] [ERROR] Activating over SSH resulted in a bad exit code: Some(1)
🚀 ℹ️ [deploy] [INFO] Revoking previous deploys
🚀 ❌ [deploy] [ERROR] Deployment to node failing-server failed, rolled back to previous generation
... ~240 seconds elapse ...
👀 ❌ [wait] [ERROR] Error waiting for activation: Timeout elapsed for confirmation
```

Total `deploy` runtime: ~250 seconds, essentially `activation_timeout` plus small overhead.

Note that the client's `🚀 ❌ [deploy] [ERROR] Activating over SSH...` line is logged immediately after activation fails. The deployer recognizes the failure quickly, but the `deploy` command does not exit until the `wait` process times out on its own.

### Root Cause

When `magic_rollback` is enabled, the deployer spawns a `wait` process on the target alongside `activate`, in a separate SSH session. `wait` watches `temp_path` via inotify for an  activation canary file, which `activate` creates only on the success path (in `activation_confirmation`). 

On failure, no canary is created and there's no other signal channel between the two processes, so `wait` blocks until its own timeout (`activation_timeout`, default 240s) elapses.

### History

Similar issue previously reported in [#58](https://github.com/serokell/deploy-rs/issues/58) (closed Feb 2021 by [#59](https://github.com/serokell/deploy-rs/issues/59)). PR #59 addressed the deployer side (recognizing activate's exit) but didn't establish a target-side signal channel between `activate` and `wait`.

paulbertin:
Fixes #372

### Description

This adds a target-side signal channel between `activate` and `wait`. On the failure path, `activate` writes a `deploy-rs-cancel-<hash>` sentinel file in `temp_path`. The `wait` process's existing inotify watcher recognizes the sentinel alongside the success canary and returns `DangerZoneError::Cancelled` immediately, instead of blocking until `activation_timeout` elapses.

### Testing

Adds unit tests and an integration test (`activation-failure-cancellation`) that deploys an  intentionally-failing profile and asserts `deploy` returns within 30s, well under the 60s timeout it is given.

Verified locally: the same test fails on `master` (asserts past the threshold, once `activation_timeout` is reached) and passes on this branch (deployment is cancelled immediately, no more `👀 ❌ [wait] [ERROR] Error waiting for activation: Timeout elapsed for confirmation`).

### DIFF
--- a/nix/tests/default.nix
+++ b/nix/tests/default.nix
@@ -20,7 +20,9 @@ let
     done <$refs
   '';
 
-  mkTest = { name ? "", user ? "root", flakes ? true, isLocal ? true, deployArgs }: let
+  mkTest = { name ? "", user ? "root", flakes ? true, isLocal ? true, deployArgs
+           , expectCancellationWithin ? null
+           }: let
     nodes = {
       server = { nodes, ... }: {
         imports = [
@@ -112,6 +114,15 @@ let
         timeout=30
       )
 
+      ${if expectCancellationWithin != null then ''
+      # Activation is expected to fail. Assert the wait process is cancelled
+      # rather than blocking until activation_timeout, so deploy returns quickly.
+      import time
+      start = time.time()
+      client.fail("deploy ${deployArgs}")
+      elapsed = time.time() - start
+      assert elapsed < ${toString expectCancellationWithin}, f"deploy took {elapsed}s, expected cancellation within ${toString expectCancellationWithin}s"
+      '' else ''
       # Make sure the hello and figlet packages are missing
       server.fail("su ${user} -l -c 'hello | figlet'")
 
@@ -120,6 +131,7 @@ let
 
       # Make sure packages are present after deployment
       server.succeed("su ${user} -l -c 'hello | figlet' >&2")
+      ''}
     '';
   };
 in {
@@ -172,6 +184,13 @@ in {
     name = "non-flake-with-flakes";
     flakes = true;
     deployArgs = "--file . --targets server";
+  };
+  # Verify activation failure triggers cancellation of the wait process,
+  # rather than waiting for the full activation timeout (default 240s).
+  activation-failure-cancellation = mkTest {
+    name = "activation-failure-cancellation";
+    deployArgs = "-s .#failing-server -- --offline";
+    expectCancellationWithin = 90;
   };
   # Pure-evaluation test for the drvPath auto-extraction. Runs without a VM.
   transform-deploy = import ./transform-deploy.nix { inherit pkgs; };
--- a/nix/tests/default.nix
+++ b/nix/tests/default.nix
@@ -119,9 +119,11 @@ let
       # rather than blocking until activation_timeout, so deploy returns quickly.
       import time
       start = time.time()
-      client.fail("deploy ${deployArgs}")
+      out = client.fail("deploy ${deployArgs} 2>&1")
       elapsed = time.time() - start
       assert elapsed < ${toString expectCancellationWithin}, f"deploy took {elapsed}s, expected cancellation within ${toString expectCancellationWithin}s"
+      # Fail on the activation error, not on some other error that happens to be quick.
+      assert "intentional activation failure" in out, f"deploy failed for an unexpected reason: {out}"
       '' else ''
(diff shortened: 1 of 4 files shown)


heitor-lassarote (changes requested):
This addresses a real, still-open issue (#372), confirmed still present on current `master`.

Two things before merge:

1. The branch is behind `master` and will need a rebase. `activation_confirmation`'s watcher setup changed as part of a race-condition fix (#385) that landed after this branch was created. The conflict is mechanical, not a design conflict with this change.

2. Testing found a race: if activation fails fast enough that the remote `activate` process writes the cancel file before `wait()`'s SSH connection is established, `wait()`'s stale-file cleanup at startup removes it before `wait()` ever sees it, and it blocks for the full `activation_timeout` instead of returning immediately.

Here's a test reproducing it, addable to the existing `mod tests` block in `activate.rs`:

```rust
fn test_closure(id: u64) -> String {
    format!("/nix/store/0000000000000000000000000000000cancel-test-{id}")
}

#[tokio::test]
async fn wait_cancelled_when_cancel_file_already_exists() {
    let id = TEMP_DIRECTORY_ID.fetch_add(1, Ordering::Relaxed);
    let temp_path = env::temp_dir().join(format!(
        "deploy-rs-cancel-before-{}-{id}",
        std::process::id()
    ));
    let closure = test_closure(id);
    fs::create_dir_all(&temp_path)
        .await
        .expect("create test directory");

    // Cancel file exists before wait() is even called.
    create_activation_cancel(&temp_path, &closure).await;

    let wait_result = wait(temp_path.clone(), closure.clone(), Some(5)).await;
    match wait_result {
        Err(WaitError::Waiting(DangerZoneError::Cancelled)) => {}
        other => panic!("expected Cancelled, got {:?}", other),
    }

    let _ = fs::remove_dir_all(&temp_path).await;
}
```

On this branch it currently returns `Err(Waiting(TimesUp))` instead of `Cancelled`.

Requesting changes for the race above. Happy to take another look once it's addressed.

paulbertin:
Thanks for the feedback @heitor-lassarote, just rebased onto `master`. 

I kept the extracted `confirmation_watcher` and threaded `WaitEvent` through it. Two extra fixes were needed because `clippy --deny warnings` and `cargo fmt --check` landed after this branch was cut.

On the race: `wait` no longer removes the sentinel at startup, so a sentinel written before its SSH connection is up is picked up by the existence check after `watcher.watch`. I included your suggested test verbatim as `wait_cancelled_when_cancel_file_already_exists` and confirmed it fails with `Waiting(TimesUp)` after 5s when the old cleanup is restored.

On cleanup: two steps now remove the sentinel file. Before activation starts, `activate` removes an old sentinel file. After `wait` detects the sentinel file, `wait` removes it. Thus no sentinel file stays on the target.

Also added unit tests for the watcher route, the canary path, and sentinel removal, and tightened the VM test. It now runs with `--activation-timeout 60` and asserts the failure is the activation error rather than only checking elapsed time.

Verified on `x86_64-linux`: the VM test fails on master's `src/` (`deploy` takes 62.5s) and passes here (2.7s).

Let me know if there's anything left you'd like to change!

heitor-lassarote (approved):
Thank you for addressing my comments and for writing tests. Your tests pass here and fail on master, so looks like this is good to go. Approving.

PR #374, opened 2026-05-05
chore: remove ill-defined escapes

bitbloxhub:
Lix was warning about these, not tested with, but should work, with upstream Nix.

In nix, "\." just escapes to "." and is not passed through literally as "\." in double-quoted strings.


Luflosi:
Duplicate of #361.

bitbloxhub:
Missed that, oops!

Issue #164, opened 2022-06-26
Incompatibility with content-addressed derivations


thufschmitt:
I've been meaning to try `deploy-rs` for my systems, but it turns out that it doesn't work if the closure to deploy contains some content-addressed derivations. For example given the following flake (slightly modified copy of the example):

<details><summary>flake.nix</summary>
<p>

```nix
# SPDX-FileCopyrightText: 2020 Serokell <https://serokell.io/>
#
# SPDX-License-Identifier: MPL-2.0

{
  description = "Deploy GNU hello to localhost";

  inputs.deploy-rs.url = "github:serokell/deploy-rs";

  outputs = { self, nixpkgs, deploy-rs }:
  let
    hello = nixpkgs.legacyPackages.x86_64-linux.runCommandNoCC "hello" {
      __contentAddressed = true;
    } ''
      mkdir -p $out/bin
      cat <<EOF > $out/bin/hello
      #!/bin/sh
      echo Hello
      EOF
    '';
  in
  {
    deploy.nodes.example = {
      hostname = "localhost";
      profiles.hello = {
        user = "balsoft";
        path = deploy-rs.lib.x86_64-linux.setActivate hello "./bin/hello";
      };
    };

    checks = builtins.mapAttrs (system: deployLib: deployLib.deployChecks self.deploy) deploy-rs.lib;
  };
}
```

</p>
</details>

I get:

```console
$ nix run github:serokell/deploy-rs .
🚀 ℹ️ [deploy] [INFO] Running checks for flake in .
warning: Git tree '/tmp/tmp.FQSqTMcjgn/deploy-rs' is dirty
warning: unknown flake output 'deploy'
🚀 ℹ️ [deploy] [INFO] Evaluating flake in .
warning: Git tree '/tmp/tmp.FQSqTMcjgn/deploy-rs' is dirty
🚀 ℹ️ [deploy] [INFO] The following profiles are going to be deployed:
[example.hello]
user = "balsoft"
ssh_user = "regnat"
path = "/1m3axr5ii129c20a74wg993p2ph1k0ivpc8siprnp3ak1pfrmavj"
hostname = "localhost"
ssh_opts = []

🚀 ❌ [deploy] [ERROR] Failed to push profile: Nix show-derivation command resulted in a bad exit code: Some(1)
```

[...]

PR #375, opened 2026-05-16
Implement support for floating output paths

poly2it:
I took a shot at resolving the issues with ca-derivations (#164) since I use them quite frequently. This resolves the issues I was having, but you might want to test the changes further. The changes were implemented with help from Claude Opus 4.7. If this is an issue, I would like to make you aware of it.

This branch keeps the user-facing API unchanged. A Nix transformation, applied in both flake and non-flake eval modes, walks `nodes.*.profiles.*`. When `path` evaluates to a derivation, it is transformed such that `path` becomes `outPath` and a sibling `drvPath` is added. The binary picks up the new `drvPath` field on `ProfileSettings` and builds the `.drv` directly. The realised output is then resolved by calling `nix path-info <drv>^out` against the local or remote store. That resolved path is used throughout copy, activate, confirm, and revoke, so no stage falls back to the eval-time placeholder.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -31,6 +31,8 @@
             "src"
             "src/bin"
             ".*\.rs$"
+            "nix"
+            "nix/.*\.nix$"
           ];
 
           cargoLock.lockFile = ./Cargo.lock;
--- a/nix/tests/default.nix
+++ b/nix/tests/default.nix
@@ -173,4 +173,6 @@ in {
     flakes = true;
     deployArgs = "--file . --targets server";
   };
+  # Pure-evaluation test for the drvPath auto-extraction. Runs without a VM.
+  transform-deploy = import ./transform-deploy.nix { inherit pkgs; };
 }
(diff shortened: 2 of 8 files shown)


poly2it:
I pushed two additional commits to resolve an issue which was only made evident after initial support was implemented.

In Nix 2.34, `path-info` on a placeholder drv looked up its hash-modulo-keyed build trace directly and found the output. I updated to the Git version of Nix 2.35 [for stability reasons](https://github.com/NixOS/nix/pull/15679), and found that the placeholder no longer has its own build-trace entry. Only the resolved derivation does, and the lookup must walk through the resolution step. The resolution does however still work through `nix build`, which is now used instead.

See this document in the Nix repository for an explanation to the changes:

https://github.com/NixOS/nix/blob/41b28ad2ed0f4dcde548fa9714136f714a7fd789/doc/manual/rl-next/build-trace-rework.md

Sereja313 (approved):

Sereja313:
Thanks!

PR #376, opened 2026-05-28
Update flake.lock and fix 403 on crate.io

alberand:
nixos-unstable has a work-around for crate.io returning 403 on the requests without user-agent. The Nix's rust package fetcher fetchCargoVendor uses python's requests library which doesn't set agent explicitly. From recently create.io returns 403 on such requests (upstream issues says that nixpkgs has exception but peoples seems to get 403 anyway). cargoLock uses uses fetchurl default curl user-agent which gets blocked by crate.io

### DIFF
--- a/flake.lock
+++ b/flake.lock
@@ -18,11 +18,11 @@
     },
     "nixpkgs": {
       "locked": {
-        "lastModified": 1743014863,
-        "narHash": "sha256-jAIUqsiN2r3hCuHji80U7NNEafpIMBXiwKlSrjWMlpg=",
+        "lastModified": 1780365719,
+        "narHash": "sha256-QfWfccTN+70ZQ4m2qlU9PiKfz2Yppq94058iJyARNwc=",
         "owner": "NixOS",
         "repo": "nixpkgs",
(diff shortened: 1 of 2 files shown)


notgne2:
I believe there's a second PR that just got merged into nixpkgs to fix cargo lock usage, with a workaround that can be applied in this flake in the meantime too. Not a fan of adding the hash

alberand:
Will you update nixpkgs, or should I update this PR? (do you have a link for that fetchurl workaround? I didn't find it)

notgne2:
https://github.com/NixOS/nixpkgs/pull/524985/changes this should be doable by putting it in `extraRegistries` instead, which can be done by using `cargoDeps = rustPlatform.importCargoLock` instead of `cargoLock.lockFile`. Not tested yet but do think this would probably fix it

alberand:
ah this one, yeah, doesn't seem to be in nixpkgs-unstable yet

alberand:
nixpkgs-unstable got updated, I updated MR as suggested and dropped flake-compact flake input update.

PR #381, opened 2026-06-16
Lint/CI gates, panic hardening, activate.profile, nullable schema

jagajaga:
A batch of small, **non-breaking** improvements. The public surface — `deploy-rs.lib.<system>.activate.*`, the `interface.json` config schema, the `deploy` CLI flags, and the `deploy`↔`activate` protocol — is unchanged.

## Commits
1. **chore: clippy lints + clippy/rustfmt/reuse checks** — resolves all clippy warnings, annotates the genuinely-many-arg functions with `#[allow]`, and adds three lint derivations to the flake `checks` (so they run in CI). The `reuse` check caught two pre-existing files missing SPDX headers (`src/command.rs`, `.github/workflows/check.yml`) — fixed.
2. **docs: example off deprecated `setActivate`** — `examples/simple` now uses `activate.custom` (the alias stays for back-compat).
3. **fix: avoid panics in the deploy waiter and flake-target parser** — the magic-rollback waiter `unwrap()`ped a oneshot `RecvError` (a successful activation finishing before the wait command returned would panic); now handled as success. The rnix fragment parser `unwrap()`ped AST nodes; malformed targets now return `ParseFlakeError` instead of panicking.
4. **feat: `activate.profile`** — a new additive activator that installs the closure into the target's nix3 `nix profile` (idempotent: replaces a previous generation of the same name). Existing activators untouched.
5. **feat: nullable optional fields in `interface.json`** — widens the schema so optional settings may be `null` (backward-compatible; also future-proofs a module-system consumer that emits explicit `null` defaults).

## Verification
Run in a `nixos/nix` (aarch64) container against a toy multi-node flake:
- `nix build .#checks.<sys>.{clippy,rustfmt,reuse,deploy-rs}` → all pass
- magic-rollback deploy → `Deployment confirmed.` (confirms the waiter fix doesn't regress the rollback path)
- `activate.profile` → installs + runs a package, idempotent on re-deploy
- `deploy` **with** schema checks and `fastConnection = null` → passes (exercises the nullable schema end-to-end)

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -184,6 +184,33 @@
 
         checks = {
           deploy-rs = self.packages.${system}.default.overrideAttrs (super: { doCheck = true; });
+
+          # Lint the Rust sources with clippy, failing on any warning.
+          clippy = self.packages.${system}.default.overrideAttrs (super: {
+            pname = "deploy-rs-clippy";
+            nativeBuildInputs = (super.nativeBuildInputs or [ ]) ++ [ pkgs.clippy ];
+            buildPhase = "cargo clippy --all-targets -- --deny warnings";
+            doCheck = false;
+            installPhase = "touch $out";
+          });
+
+          # Enforce `cargo fmt` formatting.
+          rustfmt = pkgs.runCommandLocal "deploy-rs-rustfmt" {
+            nativeBuildInputs = [ pkgs.cargo pkgs.rustfmt ];
+          } ''
+            export HOME="$TMPDIR"
+            cd ${self.packages.${system}.default.src}
+            cargo fmt --check
+            touch $out
+          '';
+
+          # Enforce REUSE licensing compliance.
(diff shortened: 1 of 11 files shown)


flake.nix
```diff
@@ -116,6 +116,21 @@
...
 
             noop = base: custom base ":";
+
+            # Install a package into the target's nix3 profile (`nix profile`).
+            # Unlike the other activators this does not just `nix-env --set` the
+            # closure into deploy-rs' own profile, it also (re)installs `base`
+            # into the calling user's `nix profile`, replacing a previous
+            # generation with the same name. Useful for declaratively rolling
+            # out a package (or a `buildEnv`) to a machine.
+            profile = base: custom base ''
```
Sereja313, line 126: It would be nice to mention this activator in the `README`.

Sereja313 (approved):

PR #382, opened 2026-06-19
Allow using environment variables

claraphyll:
*My usecase for this:* I am using a repository shared by multiple people. My username on the host is different from the one on the server, but I also don't want to hardcode a user. Therefore, I want to set the user via a local direnv.

This pull request allows setting all config options that can be set from the command line via environment variables as well.

I used Claude Opus 4.8 to write this

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -11,7 +11,7 @@ edition = "2018"
 # See more keys and their definitions at https://doc.rust-lang.org/cargo/reference/manifest.html
 
 [dependencies]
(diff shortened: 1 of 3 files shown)


claraphyll:
i've done all cli args because i didn't want to think about which ones are reasonable to pass through cli 

i think you could argue for a lot of them, e. g. "remote build only if i'm not on a powerful machine"

i'm open to discussing which, if any, options shouldn't be settable through environment variables

PR #383, opened 2026-06-23
fix: don't set ssh user by default

### LINKED ISSUE #208: Default SSH user should not be taken from current username
This works enough of the time, but is a bit of a poor decision. We should instead leave it up to SSH to decide, so users can choose their SSH usernames in an SSH config file.

claraphyll:
Fixes #208

### DIFF
--- a/src/cli.rs
+++ b/src/cli.rs
@@ -297,7 +297,7 @@ async fn get_deployment_data(
 #[derive(Serialize)]
 struct PromptPart<'a> {
     user: &'a str,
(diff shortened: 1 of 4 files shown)


claraphyll:
I am uncertain if the "sudo if username != ssh_user" mechanism works with this correctly. e. g. if you set "root" in ssh config then it will still sudo. I don't think that's really a problem, but maybe an explicit `useSudo` option could be added. It seems preferable to having to find out the SSH user configured in `~/.ssh/config` somehow

PR #387, opened 2026-07-30
fix: stream build logs when -L is passed

Sereja313:
Problem: Recent async build feature broke -L/--print-build-logs.

Solution: When -L is set, stream each build-log line above the bar, prefixed with the derivation name like native `nix build -L`.

### DIFF
--- a/src/push.rs
+++ b/src/push.rs
@@ -241,7 +241,8 @@ const RES_BUILD_LOG_LINE: i64 = 101;
 
 /// Accumulates the aggregate progress reported by `nix --log-format internal-json`
 /// so we can render a compact `x/y built, x/y copied` message next to the spinner,
-/// mirroring nix's own progress bar.
+/// mirroring nix's own progress bar. Only used when builds run concurrently (i.e.
+/// remote builds); local builds print nix's native output directly.
 #[derive(Default)]
 struct NixProgress {
     // Map of activity id -> activity type, so `result` events can be attributed.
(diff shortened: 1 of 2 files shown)


src/push.rs
```diff
@@ -232,12 +233,31 @@ pub async fn build_profile_locally(
...
+fn drv_name(drv_path: &str) -> String {
+    let base = drv_path.rsplit('/').next().unwrap_or(drv_path);
+    let base = base.strip_suffix(".drv").unwrap_or(base);
+    let name_ver = base.split_once('-').map(|(_, r)| r).unwrap_or(base);
+    let bytes = name_ver.as_bytes();
+    let cut = (0..bytes.len())
+        .find(|&i| bytes[i] == b'-' && bytes.get(i + 1).is_some_and(u8::is_ascii_digit))
+        .unwrap_or(name_ver.len());
+    name_ver[..cut].to_string()
+}
```
quentinmit: This feels very fragile - I think this is the only place where `deploy` has to parse the `nix` command line, and then you are doing a bunch of work to recreate similar log output to Nix.

Could you maybe get 99% of the way there just by modifying `cli.rs` to not set `data.deploy_data.progressbar = Some(pb.clone());` when there are no remote builds? It looks like that would result in `nix build`'s normal log output appearing on the terminal for local builds, which already are done synchronously.

Sereja313: Fair point, I removed it everywhere except in the case of multiple remote builds.

Issue #379, opened 2026-06-09
command hangs on "🚀 ℹ️ [deploy] [INFO] Running checks for flake in ."


imran-iq:
hi,

I've been running `nix run github:serokell/deploy-rs` for a quite a while without issue, but it seems after the recent changes the binary just hangs after "running checks"

If I run the version before the most recent changes with:

```
nix run github:serokell/deploy-rs/77c906c0ba56aabdbc72041bf9111b565cdd6171
```

Then it works as expected:

```
🚀 ℹ️ [deploy] [INFO] Running checks for flake in .
warning: Git tree '/home/imran/code/media_server' is dirty
warning: unknown flake output 'deploy'
<stuff>
```

But for the newer version it doesn't even get to the "Git tree is dirty" part

heitor-lassarote:
Should hopefully be resolved by #380.

heitor-lassarote:
@imran-iq, could you check if the issue has been resolved?

imran-iq:
I can try after work, I'll let you know later today (ty for the quick turnaround)

imran-iq:
Can confirm issue is resolved. Thank you once again

PR #401, opened 2026-09-10
[#261] Demarcate output from remote server

### LINKED ISSUE #261: Request: demarkate output from server vs. local machine
I've just solved the issue I mentioned in #260, which turns out to have required a restart of the nix-daemon ... on the _server_, not my local machine. This provokes another request that I think might help anybody else who runs into issues where it's ambiguous whether an messages comes from the local or remote machine

```
error: cannot connect to socket at '/nix/var/nix/daemon-socket/socket': Connection refused
error: unexpected end-of-file
```

How about putting a prefix in front of each line of stdout that's forwarded from the remote machine to local stdout? E.g.

```
📠 error: cannot connect to socket at '/nix/var/nix/daemon-socket/socket': Connection refused
📠 error: unexpected end-of-file
```

Here I've used the fax machine emoji, thinking that kind of goes with the idea of "relocated output".

heitor-lassarote:
## Description

Prefixes each line of output forwarded from the remote server's SSH session (activate, wait-for-confirmation, revoke) with 📠, so it's unambiguous whether a given line came from the server or from the local `deploy` process itself.

Adds `--no-demarcate-output` to opt out: when set, stdout/stderr are left inherited instead of piped, so a script or tool parsing `deploy`'s output isn't broken by the new prefix.

## Related issue(s)

Resolves #261

## Notes for reviewers

- Streaming semantics are preserved: output is still forwarded line-by-line as it arrives, not buffered until the process exits (this project has been bitten by that class of bug before, see #379/#380).
- `revoke()` previously never piped its SSH child's stdout/stderr at all; this PR fixes that as part of wiring up demarcation for it.
- Covered by a unit test for `forward_reader` and new e2e tests (`prefix-demarcation`, `no-demarcate-output`, `revoke-demarcation`).
- Open question worth weighing in on: demarcation defaults to **on**, opt-out via `--no-demarcate-output`. This changes stdout/stderr byte content unconditionally, including for piped/scripted callers, not just interactive terminal use (unlike `--no-progress`, which only affects TTY output since progress bars are already suppressed on non-TTY streams). Alternatives considered: default off (opt-in, no risk of breaking existing scripts, but the feature goes mostly unused since it needs discovering), or auto-detect based on whether stdout is a TTY (matches `--no-progress`'s convention, but would also silently disable the prefix in CI log output that a human reads later, which is one of the two scenarios this feature targets). Currently shipping with the opt-out default; open to switching before merge.

### DIFF
--- a/src/deploy.rs
+++ b/src/deploy.rs
@@ -4,10 +4,13 @@
 //
 // SPDX-License-Identifier: MPL-2.0
 
-use log::{debug, info, trace};
+use log::{debug, info, trace, warn};
 use std::path::Path;
 use thiserror::Error;
-use tokio::{io::AsyncWriteExt, process::Command};
+use tokio::{
+    io::{AsyncReadExt, AsyncWriteExt},
+    process::Command,
+};
 
 use crate::{command, DeployDataDefsError, DeployDefs, ProfileInfo};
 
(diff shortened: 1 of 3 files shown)


src/deploy.rs
```diff
@@ -546,7 +631,14 @@ pub async fn deploy_profile(
         let mut ssh_wait_command = Command::new("ssh");
         ssh_wait_command
             .arg(&ssh_addr)
+            .stdout(std::process::Stdio::piped())
+            .stderr(std::process::Stdio::piped())
```
Sereja313: These lines seem redundant and should only be added to the `if demarcate_output` block below. When using the `--no-demarcate-output ` it seems that nobody is consuming them, and with a large log, they can deadlock.
```suggestion

```

Sereja313:
> Open question worth weighing in on: demarcation defaults to on, opt-out via --no-demarcate-output. This changes stdout/stderr byte content unconditionally, including for piped/scripted callers, not just interactive terminal use (unlike --no-progress, which only affects TTY output since progress bars are already suppressed on non-TTY streams). Alternatives considered: default off (opt-in, no risk of breaking existing scripts, but the feature goes mostly unused since it needs discovering), or auto-detect based on whether stdout is a TTY (matches --no-progress's convention, but would also silently disable the prefix in CI log output that a human reads later, which is one of the two scenarios this feature targets). Currently shipping with the opt-out default; open to switching before merge.

Yeah, I agree, i think this should be on by default.

src/deploy.rs
heitor-lassarote: Good catch, fixed it.

Sereja313 (approved):
LGTM

